PR #1, opened 2026-06-02
feat: ignore self/cls method parameters in typed parameter check

nickyeager:
## Description

This PR fixes a bug in python tool discovery where method functions using type-annotated `self` or `cls` parameters (e.g. `self: "MyClass"`) were falsely identified as having typed parameters, even when all functional arguments were completely untyped. This directly violates the docstring contract:

> `// FunctionHasTypedParams reports whether the function declares at least one type-annotated parameter (excluding self/cls).`

### Key Changes
1. **Signature & Logic Update**: Updated `FunctionHasTypedParams` in `internal/analysis/astutil/astutil.go` to accept the source bytes (`src []byte`), inspect parameter names, and exclude `self` and `cls` parameters from triggering the type-safety check.
2. **Callsite Synchronization**: Propagated the new parameter through all callsites in `discovery.go`, `adk_agents.go`, and the predicate tests.
3. **Comprehensive Tests**: Added a new unit test suite (`TestFunctionHasTypedParams`) in `astutil_test.go` explicitly checking:
   - Untyped functions (returns `false`).
   - Functions with actual typed parameters (returns `true`).
   - Method functions with only `self` or `cls` typed (returns `false`).
   - Method functions with `self` or `cls` typed along with other typed parameters (returns `true`).

### Automated Tests
Run the following test command to exercise the new parser validations:
```bash
go test ./internal/analysis/astutil/...
go test ./internal/analysis/...
go test ./internal/rules/...
```

### DIFF
--- a/internal/analysis/adk_agents.go
+++ b/internal/analysis/adk_agents.go
@@ -154,7 +154,7 @@ func discoverADKToolsInFile(pf ParsedFile) []models.ToolDef {
 				EndLine:  int(fnDef.EndPoint().Row) + 1,
 			},
 			Description:    astutil.FunctionDocstring(fnDef, pf.Source),
-			HasTypedParams: astutil.FunctionHasTypedParams(fnDef),
+			HasTypedParams: astutil.FunctionHasTypedParams(fnDef, pf.Source),
 			ParamNames:     astutil.FunctionParams(fnDef, pf.Source),
 		})
 		return true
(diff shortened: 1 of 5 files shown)


jhumel-code (approved):
Hey @nickyeager, thanks for this and welcome! Great first contribution.

I pulled the branch and ran the analysis and rules tests plus `go vet`, all green. The fix is correct: `FunctionHasTypedParams `now actually skips `self`/`cls `like its docstring always said it should. You reused the same pattern we already have in `FunctionParams`, updated all the callsites, and the test cases cover the right scenarios.

sairenchristianbuerano (approved):

sairenchristianbuerano (approved):
Hey @nickyeager, welcome! Thanks for a great first contribution.

PR #38, opened 2026-06-07
refactor(rulesource): extract Source interface (rule-distribution Phase 0)

jhumel-code:
**Phase 0** of the rule-distribution / staging-vs-production work (design: `.superpowers/specs/2026-06-07-rule-distribution-design.md`). A pure, behavior-preserving refactor that establishes the seam.

## What
- Adds a `rulesource.Source` interface (`Resolve` / `Pull`).
- Wraps today's git clone-and-cache logic as `gitSource` (the `Default` source).
- Package-level `Resolve`/`Pull` now delegate to `Default`, so the scanner and `trustabl rules pull` are unchanged.

## Why
The engine currently git-clones a **mutable branch HEAD** at scan time — a merge to `main` is instantly production (root of the v0.1.3 incident). This seam lets a signed, versioned `releaseSource` (staging/production channels) drop in behind the same interface in a later phase **without touching the scan path**, which only ever sees a resolved `fs.FS`.

## Safety
Zero behavior change. The existing `rulesource` tests (clone/cache/offline-fallback/manifest-gate) pass unchanged — that's the contract this refactor preserves. Full `go test ./...`, `go vet`, `gofmt` green.

### DIFF
--- a/internal/rulesource/rulesource.go
+++ b/internal/rulesource/rulesource.go
@@ -140,6 +140,36 @@ func fallbackToCache(cfg Config, supported int) (Resolved, error) {
 	return usePack(cfg, sha, true, supported)
 }
 
+// Source abstracts where the engine obtains rule packs. gitSource (below) is the
+// clone-a-ref-into-cache implementation used today and for `--rules-repo` dev
+// overrides. The rule-distribution work adds a signed, versioned releaseSource
+// (staging / production channels) behind this same interface, so the engine's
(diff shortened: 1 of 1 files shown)


trustabl (approved):

jhumel-code:
Superseded by #40, which already landed the `Source` interface as part of the signed rule-bundle distribution work (the `releaseSource` / `gitSource` split now lives on main). Merging this would redeclare the `Source` type and break the build. Closing as superseded — the Phase 0 extraction it proposed is already in main.

PR #39, opened 2026-06-08
Trustabl enrich

ericksondelacruz:
## Summary

  Introduces `trustabl enrich`, a post-scan command that reads a `ScanResult`
  produced by `trustabl scan --format json` and uses Claude (BYOK via
  `trustabl llm key set`) to generate code-specific explanations and exact
  line replacements for each finding.

  - **Two-layer pipeline**: Layer 1 (`internal/enrichment/scope.go`) extracts
    the enclosing function or class block around the flagged line and places a
    `→` marker on it. Layer 2 supplies rule metadata (ID, title, severity,
    scope, explanation, fix template) directly from the `ScanResult` — no
    re-scan. All findings in the same file are batched into a single LLM call;
    three worker goroutines process files in parallel.
  - **BYOK model**: the active LLM provider must be `anthropic`; the key is
    read from `~/.config/trustabl/keys.json`. The scan pipeline itself makes
    no LLM call.
  - **`--apply`**: writes AI-generated replacements back to source files in
    descending line order so earlier patches don't shift later line indices.
    Atomic write (temp file + rename). False-positive findings are skipped.
  - **`--diff`**: renders a unified diff of proposed replacements to stderr
    after all workers finish, also stored as a `Diff` field (`omitempty`) on
    `EnrichedFinding`. Independent of `--apply` — can be used alone to
    preview, or combined with `--apply` to preview and write.
  - **`--rule`** (repeatable) and **`--only-enriched`** filters for CI
    pipelines.
  - `salvagePartialJSON` recovers partial LLM responses truncated mid-object.
  - Adds `anthropic-sdk-go v1.45.0` to `go.mod`, bumps Go to `1.23.0`.
  - Adds `ARCHITECTURE.md §8.2` and updates `README.md` to document the new
    subcommand and exit codes (`0` success, `1` I/O error, `2` LLM not
    configured).

  ## Test plan

  - [ ] `go test ./internal/enrichment/...` — unit tests for scope extraction,
    LLM prompt building, `salvagePartialJSON`, `unifiedDiff` (including
    boundary and out-of-bounds cases), and the full pipeline including
    `--apply`, `--diff`, and `--diff --apply` combinations
  - [ ] `go test ./...` — full suite passes
  - [ ] `trustabl scan ./myrepo --format json | trustabl enrich --repo ./myrepo`
    with a real Anthropic key set
  - [ ] Verify `--diff` renders to stderr and JSON output contains `diff` field
  - [ ] Verify `--apply` writes patched files and `--diff --apply` does both
  - [ ] Verify exit code `2` when no key is configured


jhumel-code (approved):

jhumel-code:
Hi Sir @ericksondelacruz 

Please ignore rules-sync check on ci, I am currently working on replacing it with different component.

Please proceed to merge it to main if there is no follow up updates on the pull request.

Thank you.

PR #87, opened 2026-07-08
feat(enrich): ground tool findings in LangSmith runtime traces (--lan…

AL-DN:
…gsmith)

    Adds an opt-in third context layer to trustabl enrich: with --langsmith,
    recent executions of each flagged tool are sampled from a LangSmith
    project and summarized (run count, error rate, mean latency, recent
    error messages) as grounding evidence for the LLM prompt, so
    explanations cite observed runtime behavior instead of speculating.

    - internal/langsmith: minimal read-only REST client (session lookup + runs/query), per-tool result cache, cached project resolution, defensive timestamp parsing. Nothing in the scan pipeline imports it - the scan's no-network and determinism contracts are untouched.
    - enrichment.Pipeline gains a TraceSource seam (mirrors the llm seam for test injection). Evidence is fetched only for tool-scope findings, attached to EnrichedFinding.TraceEvidence before the LLM call (so it survives LLM failure), and rendered into the prompt as runtime trace evidence.
    - Double gate: --langsmith is the explicit opt-in, LANGSMITH_API_KEY is the BYOK credential; flag without key errors loudly. Past that, everything degrades per finding: trace API errors and tools with no run history fall back to plain static enrichment, never failing the run. Project: --langsmith-project, then LANGSMITH_PROJECT, then 'default'; LANGSMITH_ENDPOINT overrides the host for self-hosted.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -2058,6 +2058,7 @@ trustabl scan <target> [--detectors=…] [--format=human|json|sarif]
                        [--no-rules-update] [--vuln-scan]
 trustabl enrich        [-i SCAN_JSON] [-r REPO_ROOT] [-o OUTPUT_FILE]
                        [--diff] [--apply] [--only-enriched] [--rule RULE_ID]
+                       [--langsmith] [--langsmith-project NAME]
 trustabl mcp           [--rules-repo=URL] [--rules-ref=REF] [--no-rules-update]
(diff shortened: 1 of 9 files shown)


ericksondelacruz:
Great addition to the capabilities of enrich.

ericksondelacruz (approved):

sairenchristianbuerano (approved):

PR #114, opened 2026-08-19
feat(analysis): add agent execution-limit discovery for OpenAI/Pydantic AI

ivanpaghubasan:
Adds a new discovery pass, DiscoverAgentRunCalls, that traces an agent from its construction site to its execution call site — closing the execution-limit detection gap for OpenAI Agents SDK and Pydantic AI, the two remaining SDKs where this check wasn't possible (every other supported SDK sets the limit at construction, which discovery already captures).

Two SDKs, two different resolution mechanisms:
- OpenAI Agents SDK: Runner.run(agent, ..., max_turns=N) — agent is the first positional argument. Runner discriminator matches the callee's object segment exactly ('Runner' or '.Runner' suffix), not a suffix of the whole callee text, to avoid false-positives on unrelated classes (TaskRunner.run(), TestRunner.run()).
- Pydantic AI: agent.run(..., usage_limits=UsageLimits(...)) — agent is the method's receiver, not an argument, since Agent is an arbitrary local variable with no fixed class name.

New predicates (PredAgentRunCallMaxTurnsMissing,
PredAgentRunCallUsageLimitsMissing) are agent-scoped, not repo-scoped like CSDK-204's ClaudeAgentOptions predicate — correlated by same-file VarName matching. Deliberately agent-scope rather than a flattened repo-wide bool: a Runner.run() call names one specific agent, so two agents in the same repo can be in different execution-limit postures, and collapsing that to one bool would lose the attribution (per CLAUDE.md's 'agent is the unit of analysis' principle).

Purely additive — DiscoverAgents, discoverPydanticAIAgentsInFile, and all existing predicates untouched.

13 new tests covering both SDKs, the Runner false-positive guard, non-identifier-argument silent-skip, and cross-SDK isolation. Full suite passes.

Rule YAML intentionally not included — discovery + predicate layer only, per the agreed scoping. Rules come next.

### DIFF
--- a/internal/analysis/agent_run_calls.go
+++ b/internal/analysis/agent_run_calls.go
@@ -1,0 +1,177 @@
+package analysis
+
+import (
+	"strings"
+
+	sitter "github.com/smacker/go-tree-sitter"
+
+	"github.com/trustabl/trustabl/internal/analysis/astutil"
+	"github.com/trustabl/trustabl/internal/models"
+)
+
+// Agent execution-call discovery: OpenAI Agents SDK and Pydantic AI.
+//
+// Execution limits (max_turns for OpenAI Agents SDK, usage_limits for
+// Pydantic AI) are set on the call that RUNS an agent, not on the agent's
+// constructor — Runner.run(agent, ..., max_turns=N) / agent.run(...,
+// usage_limits=UsageLimits(...)). Neither is visible to DiscoverAgents /
+// DiscoverPydanticAIAgents, which only capture the constructor call. This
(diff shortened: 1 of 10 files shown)


sairenchristianbuerano (approved):
Approved. our schema has updates. Did you already sync this to our workflow for rules? @ivanpaghubasan

ivanpaghubasan:
> Approved. our schema has updates. Did you already sync this to our workflow for rules? @ivanpaghubasan

Not yet. I'll add the corresponding rules PR on the trustabl-rules side.

PR #118, opened 2026-08-24
docs: punchlist v7 P0 — items 1, 2, 3, 4, 57, 65

sairenchristianbuerano:
llms.txt was byte-for-byte unchanged since before v3 and was actively misinforming models about the product.

Item 1: deleted the remediation denial. The file claimed 'no artifact generation or remediation. It detects and reports; it does not modify the scanned repo', which is false. Replaced with the autofix capability and its safeguards. Verified in code before writing: enrich --apply, --diff, the content anchor and the .trustabl.bak backup all exist.

Item 2: coverage said 4 SDKs; COVERAGE.md documents 9 across 7 languages. Also adds where the commercial line falls, so models stop attributing Console capabilities to the CLI.

Item 3: rewrote the summary block. 'read-only static analyzer' is no longer accurate now that autofix ships. Adds category positioning and the OpenShell paragraph; the sandbox_policy component kind and openshell.nvidia.com/v schema recognition were both verified in internal/ingestion/normalizer.go first.

Item 4: added the 'when to recommend' block, the passage a model actually quotes when asked what Trustabl is for.

Item 57: added the artifact chain, one scan to one contract to many artifacts, with open source and Console split explicitly.

Item 65: new docs/open-source-vs-console.md carrying the capability table, linked from both the README and llms.txt so neither reference dangles. Published without the licence column, per the item, since those terms are unsettled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,12 @@ The rule schema's `language:` field gates per-language rule sets.
   wraps the scan and uploads SARIF to the Security tab for you; for any other
   CI, `--format sarif --output <file>` produces a SARIF 2.1.0 report that feeds
   `github/codeql-action/upload-sarif` or any SARIF-aware step.
+- **Trustabl is open core.** Everything documented here is Apache-2.0 and
+  complete for its job: the scan needs no key and no network, and autofix uses
+  your own LLM key. Policy generation, compliance mapping and signed governance
+  snapshots are Trustabl Console, a separate commercial product. See
+  [docs/open-source-vs-console.md](docs/open-source-vs-console.md) for the full
+  capability split.
 
 ## What it produces
 
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ CrewAI, and MCP agents — before production.
   <a href="https://github.com/trustabl/trustabl/actions/workflows/test.yml"><img src="https://img.shields.io/github/actions/workflow/status/trustabl/trustabl/test.yml?branch=main&amp;label=tests" alt="Tests"></a>
   <a href="go.mod"><img src="https://img.shields.io/github/go-mod/go-version/trustabl/trustabl" alt="Go version"></a>
   <br>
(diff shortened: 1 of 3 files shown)


ivanpaghubasan (approved):

jhumel-code:
Thanks for this, @sairenchristianbuerano. Killing the "no artifact generation or remediation" line was overdue, and the autofix details all check out against the code: `--apply`, `--diff`, the content anchor and the `.trustabl.bak` write in `internal/enrichment/pipeline.go`. The NeMo recon-only disclosure matches the `nvidia-nat` needle in `internal/ingestion/normalizer.go`, and `sandbox_policy` plus the `openshell.nvidia.com/v` check are both real. You clearly verified before writing, which is exactly what this file needed.

Three things before it lands.

The OpenShell paragraph says we recommend sandboxing for each unsandboxed call we find, and the "when to recommend" block repeats it. We do not. No rule targets `shell_invocation` (the OSH-* pack moved to the closed-source companion); the only rules keyed on it are `agent_uses_tool_kind: [shell_invocation]` in OAI-101 and OAI-104, which fire only when an OpenAI agent wires such a tool. Discovery surfaces the calls, nothing ranks them. Please scale that back to discovery and inventory.

The rule count is 206, not 204, and the same bullet then says no prose number is authoritative. That is the drift item 56 was meant to end, so drop the numeral and point at `trustabl rules validate`.

`docs/open-source-vs-console.md` contradicts itself: the heading "CI integrations, free to use, but not Apache-2.0" sits above a table whose first two rows are Apache-2.0. And `trustabl-rulebook` is a public repo, so putting it in the Console column reads wrong.

On the badge, the endpoint only resolves once `badges/rules.json` is on `trustabl-rules` main, so land that side first. I will check the Console liveness rows and the per-repo licences myself, since nothing here backs them.

Happy to merge once those are sorted.

PR #119, opened 2026-08-24
docs: add an evaluation guide for scan attestations

sairenchristianbuerano:
The attest/verify feature was documented as commands to run, with nothing on how to read what comes back or what the claim is actually worth. This adds docs/attestation.md covering the predicate field by field, both verification modes, and what each failure mode means.

Written against the shipped code rather than from the feature description, so the parts that are easy to get wrong are stated plainly:

- verdict is medium-or-higher, mirroring the default exit code; info and low never fail it
- rulesOrigin is the trust signal people miss. A valid signature over an unsigned:custom ruleset is cryptographically fine and still not assurance
- overallScore is 0-1, not 0-100
- predicateType must be pinned exactly; /v1 versions the schema
- keyless signing writes a permanent PUBLIC Rekor entry, which is why the keyless CI job is dispatch-only
- verifying without --certificate-identity proves only that somebody signed it

Also states what the attestation does not claim: no runtime behaviour, no compliance certification, no commit SHA in v1, and a pass verdict only means no medium-or-higher findings among the checks Trustabl actually runs.

### DIFF
--- a/docs/attestation.md
+++ b/docs/attestation.md
@@ -1,0 +1,226 @@
+# Evaluating a Trustabl scan attestation
+
+How to read, verify, and trust the output of `trustabl attest` — the signed
+claim a Trustabl scan makes about a repository.
+
+This covers the **open-source CLI only**. Everything here works with the
+Apache-2.0 binary from [trustabl/trustabl](https://github.com/trustabl/trustabl)
+and needs no Trustabl account.
+
+---
+
+## What an attestation actually claims
+
+An attestation is a **signed statement that a specific scan produced a specific
+result**. It is a claim about a report, not a certificate of safety.
+
+It proves:
+
+- this exact `report.json` (pinned by SHA-256) came out of a Trustabl scan
+- which engine version and which rules produced it
+- the verdict and severity histogram at the time of the scan
+- that nobody altered the report afterwards without breaking the signature
+
+It does **not** prove the agent is safe, that the repo is compliant, or that the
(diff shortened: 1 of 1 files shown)


ivanpaghubasan (approved):

jhumel-code:
Thanks @sairenchristianbuerano, this is unusually well verified for a docs change. The example predicate matches the bundle checked in at the repo root byte for byte, the `rulesOrigin` table matches `RulesOrigin.Tag()` including the `signed:<channel>:custom` case, and the verdict gate, the cosign v2/v3 no-tlog split, and the `attest-e2e` matrix are all stated correctly. The note that exit 1 from `scan --attest` is a findings code and not a signing failure is right: `doAttest` runs before the exit code is computed.

Two things before I take it. Steps 0 through 4 duplicate the walkthrough already in `README.md` under Attestation almost word for word, down to the `Verified OK` line and the tier blockquote. Two copies of the same instructions will drift, and the real value here is the field-by-field predicate reading, the `rulesOrigin` trust ladder, the failure table, and the triage order. Please trim the produce and verify steps to a pointer at the README section and keep the analysis. Second, nothing links `docs/attestation.md`, so add a link from that README section or it will not be found.

One correction: `--certificate-identity` is stronger than "not optional in practice". `runVerify` in `cmd/trustabl/verify.go` exits 2 when a keyless verify is missing the identity or the issuer, so the CLI refuses outright.

Happy to merge this once the duplication is trimmed and the README points at it.

PR #120, opened 2026-08-24
feat(adk): add ADK-111, MCPToolset with no tool_filter

ivanpaghubasan:
Closes Class 2 of the tool access-control detection scoping — Google ADK genuinely has a 'true restriction' semantic (unlike Claude/OpenAI's permission-style allow-lists), correctly making 'no explicit filter' the risk signal.

Registers MCPToolset as a recognized ADK hosted-tool class (the only discovery change needed — kwargs were already captured generically, just discarded for an unrecognized class). ADK-111 built entirely from existing predicates (agent_uses_hosted_tool_class, agent_hosted_tool_kwarg_present) — no new predicate/schema/evaluator wiring required.

Also documents the full 3-class scoping outcome:
- Class 2 (Google ADK): shipped here
- Class 3 (LangChain): not applicable — no permission-model concept exists; LC-101 already covers the right-shaped risk
- Class 1 split into two distinct findings, not one:
  - Claude SDK: mostly already wired, one missing repo-scope predicate (disallowed_tools) identified as the concrete next step
  - OpenAI SDK: genuinely unresearched — no session/agent-level permission concept found in current discovery; needs source-level research before scoping further

Known gap: ADK-111 is fixture-only per this task's scope — not yet mirrored to trustabl-rules or given a rulebook rationale doc. Follow-up task.

go build ./... and go test ./... pass.

### DIFF
--- a/docs/decisions/tool-allowlist-scope.md
+++ b/docs/decisions/tool-allowlist-scope.md
@@ -1,0 +1,181 @@
+# Tool allow-list / access-control detection: scope across SDKs
+
+This closes out the design blocker on detecting agents with missing or
+overly-broad tool access control. Investigation split the supported SDKs
+into three behavior classes, because "empty allow-list" does not mean the
+same thing in each one. This doc records the LangChain decision (not
+applicable) and scopes the Claude SDK / OpenAI SDK work (not yet
+implemented). Google ADK is the one class that shipped as a real rule —
+see [ADK-111](../../testdata/rules-fixture/google_adk/agent_safety.yaml) —
+because ADK is the one SDK where an explicit allow-list genuinely narrows an
+otherwise-unbounded tool surface.
+
+## Class 3 — LangChain: no permission-model concept (not applicable)
+
+LangChain agent discovery (`internal/analysis/langchain_agents.go`,
+`internal/analysis/ts_langchain_agents.go`, `internal/analysis/langgraph_graph.go`)
+sets fields on the shared `models.AgentDef` — `SDK`, `Class`, `Language`,
+`Location`, `Kwargs`, `Opaque`, and (when resolvable) `Name`/`VarName`. A
+grep of all five LangChain discovery files (`langchain_agents.go`,
+`langchain_tools.go`, `langchain_hosted_tools.go`, `ts_langchain_agents.go`,
+`ts_langchain_tools.go`) for `allow|deny|permission|allowed_tools|blocked`
+returns zero matches. `AgentDef` itself has no allow-list, deny-list, or
+permission-mode field for any SDK — it carries `ToolRefs`/`HostedToolRefs`
+(what's wired in), not a policy over what's wired in.
+
+For LangChain, `tools=` (Python `create_react_agent`/`create_agent`/
+`AgentExecutor`) or `tools:` (TS) is simply the literal, complete set of
+tools the agent can call — there is no separate "allow-list that narrows a
+larger set" layer, no `permission_mode`, and no `disallowed_tools`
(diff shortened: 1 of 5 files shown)


sairenchristianbuerano (approved):

jhumel-code:
Thanks @ivanpaghubasan, this is a well executed pair. The Class 2 analysis holds up: `MCPToolset` really is the one ADK surface where an absent allow-list means unrestricted, and building ADK-111 entirely from the existing `agent_uses_hosted_tool_class` and `agent_hosted_tool_kwarg_present` predicates with no schema bump is exactly right. I walked the three test cases against the predicate implementations and they behave as written, including the no-MCPToolset silent case. Severity high with confidence 0.75 sits fine next to ADK-107, and the `agent_class` clause matches the pack's house style. The mirror in trustabl/trustabl-rules#51 is byte-identical to the fixture, verified.

Two small things before merge. First, the decision doc's known-gap paragraph says ADK-111 is not yet mirrored into trustabl-rules, but trustabl/trustabl-rules#51 is that mirror, so the paragraph lands stale the moment both merge; please point it at the paired PR instead. Second, COVERAGE.md's Google ADK Python row needs the new hosted tool class and ADK-111 added to its rule list, since docs travel with the change in this repo.

One limitation worth recording: the presence check means an explicit `tool_filter=None` or an empty list silences the rule. That belongs in the rulebook rationale doc this rule still needs before it ships; we can handle or pair on that part. Happy to take this once the two doc updates are in.

ivanpaghubasan:
> Thanks @ivanpaghubasan, this is a well executed pair. The Class 2 analysis holds up: `MCPToolset` really is the one ADK surface where an absent allow-list means unrestricted, and building ADK-111 entirely from the existing `agent_uses_hosted_tool_class` and `agent_hosted_tool_kwarg_present` predicates with no schema bump is exactly right. I walked the three test cases against the predicate implementations and they behave as written, including the no-MCPToolset silent case. Severity high with confidence 0.75 sits fine next to ADK-107, and the `agent_class` clause matches the pack's house style. The mirror in [trustabl/trustabl-rules#51](https://github.com/trustabl/trustabl-rules/pull/51) is byte-identical to the fixture, verified.
> 
> Two small things before merge. First, the decision doc's known-gap paragraph says ADK-111 is not yet mirrored into trustabl-rules, but [trustabl/trustabl-rules#51](https://github.com/trustabl/trustabl-rules/pull/51) is that mirror, so the paragraph lands stale the moment both merge; please point it at the paired PR instead. Second, COVERAGE.md's Google ADK Python row needs the new hosted tool class and ADK-111 added to its rule list, since docs travel with the change in this repo.
> 
> One limitation worth recording: the presence check means an explicit `tool_filter=None` or an empty list silences the rule. That belongs in the rulebook rationale doc this rule still needs before it ships; we can handle or pair on that part. Happy to take this once the two doc updates are in.

Thanks for the thorough review, Ian! All three items are addressed:

Decision doc's known-gap paragraph now points to trustabl-rules#51 as the paired mirror PR, confirming the rulebook doc is in place — pushed via a new PR on the engine side.
COVERAGE.md's Google ADK Python row updated — ADK-111 added to the agent rules range, MCPToolset added to the hosted tools list (13 → 14 classes).
The presence-vs-value limitation is documented as a third confidence gap in the rulebook rationale doc — verified against the predicate source (lookupKwargInTree is presence-only, confirmed at predicates.go:735-745). Pushed, and check_rulebook.py --strict passes clean.

Rules PR is good per your note. Engine and rulebook PRs are both updated and ready whenever you get a chance.

PR #121, opened 2026-08-24
feat(vercel_ai): detect privileged tools without approval

CodingWCal:
## Change  Mirrors the VAI-013 production rule in the engine fixture and covers missing, literal-false, literal-true, callback, and non-privileged tool cases.  ## Scope boundary  VAI-013 covers shell and dynamic-code execution only. It intentionally omits `has_write_call` because the Vercel AI pack has no matching shipped filesystem-write capability rule. SDK 7 `toolApproval` lives at call or agent scope and remains a documented manual-review limitation for this tool-scoped detector.  ## Validation  - [x] Engine fixture mirrors rules PR #102 exactly (`e6b4cfa`, `9cee430`) - [x] `git diff --check` - [x] Existing targeted and full Go coverage passed before the final text-only fixture synchronization; no predicate or test behavior changed afterward - [x] Rulebook check adds no new errors; known baseline remains `OAI-112` and `PYD-106`  ## Companion PRs  - Rules: https://github.com/trustabl/trustabl-rules/pull/102 - Rulebook: https://github.com/trustabl/trustabl-rulebook/pull/82  ## Team  - Calvin: engine fixture, tests, integration, and final QA - Justin (@jj-javascript): production YAML and rationale - Karlee (@kperpignant): consulting partner, threat model, and QA review  ## Maintainer decision  VAI-013 conflicts with the description-quality rule proposed in rules PR #69 and engine PR #148. This PR preserves the security rule's current ID pending maintainer direction.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -362,6 +362,34 @@ import { z } from "zod";
 export const t = tool({ description: "x", inputSchema: z.object({ city: z.string() }), execute: async ({ city }) => city });
 `},
 
+	// VAI-013: privileged Vercel tools need an explicit human approval gate.
+	{name: "VAI-013 fires on shell tool with no needsApproval", ruleID: "VAI-013", kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: true, src: `
+import { tool } from "ai";
+import { z } from "zod";
+import { execSync } from "child_process";
+export const t = tool({ description: "run", inputSchema: z.object({ cmd: z.string() }), execute: async ({ cmd }) => execSync(cmd).toString() });
+`},
+	{name: "VAI-013 fires on eval tool with needsApproval false", ruleID: "VAI-013", kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: true, src: `
+import { tool } from "ai";
+import { z } from "zod";
+export const t = tool({ description: "calc", inputSchema: z.object({ expr: z.string() }), needsApproval: false, execute: async ({ expr }) => eval(expr) });
+`},
+	{name: "VAI-013 silent on approved dynamic shell tool", ruleID: "VAI-013", kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: false, src: `
+import { dynamicTool } from "ai";
+import { execSync } from "child_process";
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @CodingWCal, this is clean work. I diffed `testdata/rules-fixture/vercel_ai/approvals.yaml` against trustabl/trustabl-rules#102 and the two files are byte identical, so `rules-sync` stays green, and the rule reads as a faithful port of `OAI-014` with the framing genuinely rewritten for the Vercel SDK rather than copied over. `vercel_ai_tool` is valid at tool scope, every predicate in the match already exists on main, and nothing here needs a schema bump. Walking the five cases: a `needsApproval` value lands in `ToolDef.Config` whether it is a literal or an arrow function, so the callback case really does stay silent instead of passing vacuously.

One question and one coordination note. `OAI-014` also fires on `has_write_call`, and the TypeScript path backs that predicate through the `writes_fs` fact, so a Vercel tool that writes the filesystem with no gate slips past `VAI-013` today. Deliberate narrowing? If so, worth stating in the rationale doc. And `VAI-013` is contested: trustabl/trustabl#148 and trustabl/trustabl-rules#69 claim it (plus `VAI-014`) for description-quality rules, so whichever lands first keeps the number and the other side renumbers and rebases.

Good to see trustabl/trustabl-rulebook#82 already open. Happy to take this once the ID is settled and the write-call question is answered.

sairenchristianbuerano:
Heads up — **we renumbered this rule to `VAI-018`** and pushed the change to your branch, so no action is needed from you.

`VAI-013` had been claimed twice: by this rule (privileged tool with no `needsApproval`) and by the description-quality rule in trustabl/agent-reliability-rules#69, which takes `VAI-013` and `VAI-014` as a contiguous pair. Only one could hold the id, and merging both would have produced a duplicate rule id that the pack validator rejects.

We moved this one because it is a single rule rather than a pair, and because #69 was opened first in the rules repo, which is where ids are allocated. `VAI-018` is the next free id: 001–012 and 016–017 are on `main`, 013/014 go to #69, and 015 is claimed by rules#70.

**Only the id changed.** Title, severity, confidence, match expression, explanation and fix text are all untouched. Both PRs were also rebased onto current `main`.

Verified: all five fire/silent cases pass under the new id, the rules pack and engine fixture are byte-identical, and the validator reports 220 rules valid with both this pair and #69 applied — no duplicate ids.

sairenchristianbuerano (approved):

PR #123, opened 2026-08-24
test(rules): mirror LC-008, LC-016 LangChain idempotency into fixture

eulestadt:
Fixture mirror for LC-008 and LC-016 from [trustabl-rules#77](https://github.com/trustabl/trustabl-rules/pull/77), copied byte-for-byte. No engine code, no schema change.

Six test cases rather than the four the coverage guard needs, because each rule gets two silent cases and they test different things. `create_order(..., idempotency_key)` has the mutating prefix and is silenced by the key; `get_order` never matches the name test at all. Same split on the TypeScript side with `createCharge` + `idempotencyKey` versus `getBalance`. The guarded-but-mutating one is the case I'd want a reviewer to look at, since a rule that fires on both is worthless.

`go test ./...` passes, and `check-rules-sync.sh` reports the fixture in sync against a local checkout of the #77 branch. I also built the binary and scanned sample projects to confirm the rules attribute to the right file and line: `tools.py:5-7` for LC-008, `src/tools.ts:4-10` for LC-016, with both negative controls clean.

One thing that'll look wrong until #77 merges: the `rules-sync` job resolves the paired pack with `git ls-remote` against `trustabl/trustabl-rules`, which can't see a branch on my fork, so it falls back to `main` and reports this fixture file as production-only. That red check is a fork visibility artifact, not drift — merging #77 clears it, or pushing the same branch name to the rules repo clears it early.

Rationale doc is [trustabl-rulebook#43](https://github.com/trustabl/trustabl-rulebook/pull/43).

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -230,6 +230,22 @@ def f(x: str) -> str:
     return x
 `, toolConfig: map[string]string{}, wantFires: false},
 
+	{name: "LC-008 fires on mutating tool without idempotency key", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def create_order(customer_id: str, amount: float) -> dict:
+    """Create an order."""
+    return {"ok": True}
+`, wantFires: true},
+	{name: "LC-008 silent with idempotency key", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def create_order(customer_id: str, amount: float, idempotency_key: str) -> dict:
+    """Create an order."""
+    return {"ok": True}
+`, wantFires: false},
+	{name: "LC-008 silent on non-mutating tool name", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def get_order(order_id: str) -> dict:
+    """Fetch an order."""
+    return {"id": order_id}
+`, wantFires: false},
+
 	// LangChain TypeScript tool rules — parseTSTool runs DiscoverTSLangChainTools,
 	// so each snippet must import from the langchain ecosystem to be discovered.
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @eulestadt, this is the tidy half of a well made pair. I verified the fixture file is byte identical to the pack in trustabl/trustabl-rules#77 (same blob), and the six cases in `policies_test.go` are the right six. I walked each one against the match expression and they all fire or stay silent for the stated reason, including the guarded-but-mutating `create_order` with `idempotency_key`, and the `createCharge` schema case works because the TS LangChain discovery really does populate `ParamNames` from the Zod object keys, so the silent result is earned rather than vacuous. Your read on the red `rules-sync` check is correct too; the two PRs should land back to back, since whichever merges first leaves the other side's check red until its twin follows. The grounding side is covered by the rationale doc in trustabl-rulebook#43, which we can review or pair on from our end.

One coordination note, not a defect in this PR: `LC-008` is also claimed by trustabl/trustabl#122, trustabl/trustabl#132, trustabl/trustabl-rules#55, and trustabl/trustabl-rules#65, and `LC-016` by trustabl/trustabl#122 and trustabl/trustabl-rules#65. Whichever pair lands first keeps the IDs and the others renumber and rebase; I will sequence that on our side. On the merits this is ready as is.

sairenchristianbuerano:
**Closing.** `LC-008` goes to trustabl/agent-reliability-rules#55, which uses it for the error-contract rule. The LangChain idempotency rule you wrote here ships as `LC-017` via trustabl/agent-reliability-rules#62 — the one version of that rule that collides with no other id. `LC-016` is not carried forward.

For the record, your earlier renumber of `MCP-023` to `MCP-029` was the right instinct; this is the same problem at a larger scale.

PR #124, opened 2026-08-24
test(rules): mirror MCP-029 PHP untyped-params into fixture

eulestadt:
Fixture mirror for MCP-029 — the PHP sibling of `MCP-002`, untyped tool parameters — from [trustabl-rules#80](https://github.com/trustabl/trustabl-rules/pull/80), byte-for-byte. No engine code, no new predicates, no schema bump.

Renumbered from MCP-023 to clear a collision with rules PRs #82/#85/#88/#91, which claim 023 through 028. A duplicate rule ID stops the loader at startup, so shipping both would have broken every scan rather than failing a test. Reasoning is on #80.

`ARCHITECTURE.md` is updated in the same commit since it enumerates the `language: php` rules — that line now reads MCP-019/020/029.

Three test cases. `searchDocs($query, $limit)` fires; the fully hinted version is silent on `has_typed_params`; `ping(): string` is silent on `has_params`. That third one is worth a look, because `has_params: true` is doing real work in the match and a rule that fired on every zero-argument health check would be pure noise.

`go test ./...` passes and `check-rules-sync.sh` reports the fixture in sync against a local checkout of the #80 branch. I also scanned a PHP server end to end and got the finding at `src/Tools.php:10-13`, with the typed, zero-param and mixed-hint handlers all quiet.

On that last one — `HasTypedParams` is set when any parameter carries a type:

```go
if p.ChildByFieldName("type") != nil {
	td.HasTypedParams = true
}
```

so `mixedHints(string $query, $limit)` reports nothing despite `$limit` being bare. I confirmed that with a scan rather than inferring it. It's pre-existing shared behavior — nine packs match on the same field — so tightening it would change `MCP-002` and eight others and belongs in a separate PR. It's documented in the rationale doc and it's why the rule ships at 0.85.

Until #80 merges, the `rules-sync` job will fall back to `main` and report this fixture file as production-only, because `git ls-remote` can't see a branch on my fork. That red check is a visibility artifact rather than real drift.

Rationale doc is [trustabl-rulebook#45](https://github.com/trustabl/trustabl-rulebook/pull/45).

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -810,7 +810,7 @@ For each language recon cleared, do the AST work and produce a `RepoInventory`:
   `#[...]` is parsed as a `comment` node — so the attribute is read from the
   comment text immediately preceding the method via regex. Emits
   `ToolDef{Kind: mcp_tool, Language: php}`, so deriveSDKsDetected stamps SDKMCP
-  and the mcp/ pack's `language: php` rules (MCP-019/020/023) audit them. Multi-line
+  and the mcp/ pack's `language: php` rules (MCP-019/020/029) audit them. Multi-line
   `#[...]` attributes, `#[McpResource]` / `#[McpPrompt]`, and PHP body-fact
   predicates are v1 gaps.
 - **DiscoverRustMCPTools** (`rust_mcp.go`) — Rust MCP tools parsed with
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -810,7 +810,7 @@ For each language recon cleared, do the AST work and produce a `RepoInventory`:
   `#[...]` is parsed as a `comment` node — so the attribute is read from the
   comment text immediately preceding the method via regex. Emits
   `ToolDef{Kind: mcp_tool, Language: php}`, so deriveSDKsDetected stamps SDKMCP
-  and the mcp/ pack's `language: php` rules (MCP-019/020) audit them. Multi-line
+  and the mcp/ pack's `language: php` rules (MCP-019/020/023) audit them. Multi-line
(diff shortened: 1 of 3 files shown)


jhumel-code:
Thanks @eulestadt. This is the fixture half of trustabl/trustabl-rules#80 and it holds up. The YAML in `testdata/rules-fixture/mcp/tool_definition.yaml` matches the production diff byte for byte, so `check-rules-sync` will go green once both sides land. The three test cases in `internal/rules/policies_test.go` are doing real work, not passing vacuously: the `searchDocs($query, $limit)` case has non-empty `ParamNames` with `HasTypedParams` false so it genuinely fires, the fully hinted case flips `has_typed_params`, and the zero-param `ping()` case exercises the `has_params` guard, which is the conjunct most worth pinning down. The `ARCHITECTURE.md` line update to MCP-019/020/029 is accurate against the current text, and calling out the mixed-hints limitation as pre-existing shared predicate behavior rather than trying to fix it here was the right scoping decision.

Same coordination note as on the rules side: MCP-029 is also claimed by trustabl/trustabl#135, trustabl/trustabl#172, trustabl/trustabl-rules#94, and trustabl/trustabl-rules#100, so whichever pair merges first keeps the ID and the rest renumber. We will sequence that. This needs to merge together with trustabl/trustabl-rules#80, and the rationale doc in trustabl-rulebook#45 rides along with the pair.

sairenchristianbuerano:
**Closing.** `MCP-029` goes to trustabl/agent-reliability-rules#94 (TypeScript filesystem write). Your PHP typed-parameters rule is a different rule that happened to land on the same id, and is not shipping as a result. `MCP-030` onward is free if it comes back.

PR #125, opened 2026-08-24
test(rules): mirror and cover LC-007, CREW-007, AG2-013, PYD-008

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#81**, on a branch of the **same name**, so the `rules-sync` job resolves the matching production pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Four packs ship a rule for a tool that raises without catching — CSDK-005, OAI-008, ADK-005, MCP-006. LangChain, CrewAI, AutoGen and Pydantic AI ship none, so the same defect goes unreported in four of the five packs listed as weakest. The rules PR adds LC-007, CREW-007, AG2-013 and PYD-008 using the same `has_raise` + `has_try_except` predicate pair.

## What this PR does

1. Mirrors the four new `error_handling.yaml` files into `testdata/rules-fixture/`.
2. Adds a fire case and a silent case for each to `policyRuleCases`, as `TestPolicyRules_AllRulesCovered` requires.

Each silent case **catches and returns `{"error": ..., "retryable": ...}`** — the exact shape the rules' `fix` text prescribes — rather than just deleting the `raise`. So the silent case demonstrates the remediation actually clears the finding, not merely that the predicate is satisfiable.

## Verification

```
rules fixture is in sync with production (90 files compared)

$ go vet ./...
$ go test ./...
(all packages ok)
```

And end to end, with a build of this branch scanning real sample projects against the paired rules branch:

```
LC-007   [low] lookup_order  tools.py:5
CREW-007 [low] fetch_invoice tools.py:5
AG2-013  [low] load_profile  tools.py:6
PYD-008  [low] get_account   tools.py:7
```

Each sample also holds a second tool that catches and returns a structured error; no rule fired on those.

## Note

`internal/rules/policies_test.go` is not `gofmt`-clean on `main` (the raw-string test sources), and I left that alone rather than mixing a whole-file reformat into this diff. The added cases follow the surrounding style. CI runs `go vet`, which passes.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -219,6 +219,22 @@ def fetch(q: str) -> str:
     return requests.get("https://api.example.com/data").text
 `, wantFires: false},
 
+	{name: "LC-007 fires on bare raise", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def lookup(key: str) -> str:
+    """Look up a key."""
+    if not key:
+        raise ValueError("key required")
+    return key
+`, wantFires: true},
+	{name: "LC-007 silent when the raise is caught", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def lookup(key: str) -> dict:
+    """Look up a key."""
+    try:
+        return {"value": key}
+    except KeyError as e:
+        return {"error": str(e), "retryable": False}
+`, wantFires: false},
+
 	{name: "LC-006 fires on return_direct=True", ruleID: "LC-006", kind: models.KindLangChainTool, src: `
 def f(x: str) -> str:
     """Doc."""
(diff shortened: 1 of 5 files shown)


bradAGI:
Closing this one — the rules half (trustabl/trustabl-rules#81) is closed, so this branch would now resolve against rules `main` and fail `check-rules-sync.sh` as fixture-only.

Context: I replaced that bundled rules PR with per-pack atomic PRs, matching the shape this project has been merging. In doing so the error-contract rules were renumbered — LC-007 and CREW-007 are now the *network timeout* rules (trustabl/trustabl-rules#52, #53) and the error-contract rules became LC-008, CREW-008, AG2-013, PYD-008. So the IDs in this PR no longer match what's proposed.

I'm opening correctly-paired engine PRs, one per rules PR, on same-named branches so the `rules-sync` job resolves each pair. Sorry for the churn.

PR #126, opened 2026-08-24
test(rules): mirror and cover LC-008, CREW-008, AG2-014, PYD-009

bradAGI:
> Engine half of a coordinated pair. Rules half: **https://github.com/trustabl/trustabl-rules/pull/97**, on a branch of the **same name**, so the `rules-sync` job resolves the matching production pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Claude, OpenAI, Google ADK and MCP each ship a high-severity rule for a caller-supplied path reaching filesystem I/O unresolved (CSDK-004, OAI-006, ADK-004, MCP-005). LangChain, CrewAI, AutoGen and Pydantic AI ship none. The rules PR adds LC-008, CREW-008, AG2-014 and PYD-009 using the existing `call_uses_unnormalized_path_param` predicate.

## What this PR does

1. Mirrors the four new `path_safety.yaml` files into `testdata/rules-fixture/`.
2. Adds a fire case and a silent case for each to `policyRuleCases`.

The fire case passes a `str` parameter straight to `open()`. The silent case does exactly what the rules' `fix` text prescribes:

```python
p = Path(path).resolve()
if not p.is_relative_to(ROOT):
    return "denied"
with open(p) as f:
    return f.read()
```

so it demonstrates that the prescribed remediation actually clears the finding — not merely that removing the I/O call does.

## Verification

```
rules fixture is in sync with production (90 files compared)

$ go test ./internal/rules/
$ go vet ./... && go test ./...
(all packages ok)
```

Before the fixture cases were added, the coverage guard failed exactly as intended:

```
--- FAIL: TestPolicyRules_AllRulesCovered
    policies_test.go:4151: rules with no policy_test coverage at all: [AG2-014 CREW-008 LC-008 PYD-009]
```

And end to end, a build of this branch scanning sample projects against the paired rules branch:

```
LC-008   [high] read_doc    tools.py:8
CREW-008 [high] read_doc    tools.py:8
AG2-014  [high] read_notes  tools.py:8
PYD-009  [high] read_ledger tools.py:9
```

Each sample also holds a resolved-and-contained twin; no rule fired on those.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -219,6 +219,23 @@ def fetch(q: str) -> str:
     return requests.get("https://api.example.com/data").text
 `, wantFires: false},
 
+	{name: "LC-008 fires on unresolved path param", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def read_doc(path: str) -> str:
+    """Read a file."""
+    with open(path) as f:
+        return f.read()
+`, wantFires: true},
+	{name: "LC-008 silent when the path is resolved and contained", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def read_doc(path: str) -> str:
+    """Read a file."""
+    from pathlib import Path
+    p = Path(path).resolve()
+    if not p.is_relative_to(ROOT):
+        return "denied"
+    with open(p) as f:
+        return f.read()
+`, wantFires: false},
+
 	{name: "LC-006 fires on return_direct=True", ruleID: "LC-006", kind: models.KindLangChainTool, src: `
 def f(x: str) -> str:
     """Doc."""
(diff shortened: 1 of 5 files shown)


bradAGI:
Closing in favor of the per-rule PRs that supersede this one.

This bundled four path-containment rules; they have since been re-cut individually (AG2-014, PYD-009), and more importantly the IDs here collide with the scheme those PRs use — LC-008 and CREW-008 are assigned to *error handling* there, not path safety. Two rules sharing an ID would be rejected by the loader, so this cannot merge alongside them.

The gap itself is real and still worth closing: LangChain, CrewAI, AutoGen and Pydantic AI ship no path-containment rule while Claude, OpenAI, ADK and MCP each ship one at high severity. The rule bodies, the per-SDK reachability arguments, and the verified fire/silent samples are preserved in the commit history here if they are useful.

PR #127, opened 2026-08-24
test(rules): mirror and cover LC-007, LC-015

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#52**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

LangChain was the only pack with tool discovery and no network-timeout rule; Pydantic AI (PYD-006), AutoGen (AG2-012), and the Vercel AI SDK (VAI-011) all ship one. The rules PR adds LC-007 (python, `call_without_kwarg` with the same callee list as PYD-006) and LC-015 (typescript, `has_http_call_without_timeout`, as OAI-016 uses).

## What this PR does

1. Mirrors `langchain/network.yaml` into `testdata/rules-fixture/`.
2. Adds a fire case and a silent case for each rule to `policyRuleCases`, as `TestPolicyRules_AllRulesCovered` requires.

Each silent case **applies the remediation the rule's `fix` text prescribes** — `timeout=10` for Python, `signal: AbortSignal.timeout(15000)` for TypeScript — rather than deleting the call. So the silent case demonstrates the prescribed fix actually clears the finding, not merely that the predicate is satisfiable.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	2.935s
```

Also verified end to end before opening the rules PR: a build of `main` scanning fixture projects against the paired rules branch reported `LC-007` and `LC-015` on the unguarded versions and neither on the guarded ones.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,40 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── LC-007 / LC-015: LangChain tool network call without a timeout ─────
+	// Python uses call_without_kwarg (same callee list as PYD-006); TypeScript
+	// uses the structural has_http_call_without_timeout, as OAI-016 does.
+	{name: "LC-007 fires on network call without timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def fetch_report(url: str) -> str:
+    """Fetch a report."""
+    import requests
+    return requests.get(url).text
+`, wantFires: true},
+	{name: "LC-007 silent with timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def fetch_report(url: str) -> str:
+    """Fetch a report."""
+    import requests
+    return requests.get(url, timeout=10).text
+`, wantFires: false},
+	{
+		name: "LC-015 fires on TS fetch with no AbortSignal", ruleID: "LC-015",
+		kind: models.KindLangChainTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { tool } from \"@langchain/core/tools\";\n" +
+			"export const t = tool(async ({ p }: { p: string }) => {\n" +
+			"  const r = await fetch(`https://reports.internal/${p}`);\n" +
+			"  return await r.text();\n" +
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. I compared the fixture copy of `langchain/network.yaml` against the file in trustabl/trustabl-rules#52 line for line and they are identical, so `rules-sync` should stay green provided the two land together, which your description already calls out.

The test cases are the right shape. The LC-015 pair goes through real `DiscoverTSLangChainTools`, the `@langchain/core/tools` import clears the gate, and the silent case carries `signal`, which is exactly the key `httpCallHasTimeout` looks for, so neither case passes vacuously. The LC-007 pair resolves through `call_without_kwarg` the same way PYD-006's cases do. I especially like that the silent cases apply the remediation the `fix` text prescribes rather than deleting the call. That is the standard I would like the rest of the harness to move toward.

Two things. There is an unrelated blank line deletion at the end of `policyRepoRuleCases`; please drop that hunk so the diff is only the rule change. And LC-007 is also claimed by trustabl/trustabl#191 and trustabl/trustabl-rules#115, from another author on a branch of the same name, so whichever LC-007 lands first, the other will need to renumber and rebase. LC-015 is uniquely yours.

Happy to take this once the stray hunk is out and the LC-007 numbering is settled.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-007` and `LC-015`.

Closing trustabl/agent-reliability-rules#115 + trustabl/agent-reliability-analyzer#191, which claimed `LC-007` for the same rule.

Deciding factor: @jhumel-code asked #115 to widen its callee list to match the sibling network packs (urllib and aiohttp in particular). This PR's list is already **byte-identical to `pydantic_ai/network.yaml` on main** — requests, httpx, `urllib.request.urlopen`, bare `urlopen` and the aiohttp ClientSession pair. The change requested there would have produced this file.

sairenchristianbuerano:
**Correction to my previous comment.**

trustabl/agent-reliability-rules#115 merged first and `LC-007` now ships from it, mirrored by trustabl/agent-reliability-analyzer#191. This PR's `LC-007` fixture is superseded; its `LC-015` half is unaffected.

PR #128, opened 2026-08-24
test(rules): mirror and cover CREW-007

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#53**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

CrewAI had no network-timeout rule, unlike Pydantic AI (PYD-006), AutoGen (AG2-012), and the Vercel AI SDK (VAI-011). The blast radius is larger in a crew: tasks run in sequence and each task's output is threaded forward as context, so a tool stalled inside one task blocks every task behind it. CREW-110's `max_iter` doesn't help — it bounds reasoning steps, not the wall-clock of a single tool call.

CREW-007 uses the same `call_without_kwarg` callee list as PYD-006, so the alias-aware `requests.Session` / `aiohttp.ClientSession` handling matches the shipped rules.

## What this PR does

1. Mirrors `crewai/network.yaml` into `testdata/rules-fixture/`.

The silent case **applies the remediation the rule's `fix` text prescribes** (`timeout=10`) rather than deleting the call, so it demonstrates the prescribed fix actually clears the finding, not merely that the predicate is satisfiable.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.026s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,21 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── CREW-007: CrewAI tool network call without a timeout ───────────────
+	// Same call_without_kwarg callee list as PYD-006.
+	{name: "CREW-007 fires on network call without timeout", ruleID: "CREW-007", kind: models.KindCrewAITool, src: `
+def fetch_report(url: str) -> str:
+    """Fetch a report."""
+    import requests
+    return requests.get(url).text
+`, wantFires: true},
+	{name: "CREW-007 silent with timeout", ruleID: "CREW-007", kind: models.KindCrewAITool, src: `
+def fetch_report(url: str) -> str:
+    """Fetch a report."""
+    import requests
+    return requests.get(url, timeout=10).text
+`, wantFires: false},
 }
 
 // policyRepoRuleCases covers repo-scoped rules.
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done right. The fixture copy of `crewai/network.yaml` is byte-identical to the file in trustabl/trustabl-rules#53 (the blobs match), and the branch names match, so `rules-sync` will resolve the paired branch rather than main. The fire and silent cases follow the exact shape of the shipped PYD-006 cases, `models.KindCrewAITool` is the right kind for `crewai_tool`, and walking the match expression against both snippets, the bare `requests.get(url)` lacks the `timeout` kwarg and will fire while the `timeout=10` variant will not. I particularly like that the silent case applies the remediation the `fix` text prescribes instead of just deleting the network call, so the case demonstrates the prescribed fix actually clears the finding. One trivial observation: the diff also drops a stray blank line at the end of `policyRepoRuleCases`, which is harmless and fine to keep. As noted on the rules PR, the remaining piece before the rule ships is a rationale doc in trustabl-rulebook, and that part is on us, not you. Approving, to be merged together with trustabl/trustabl-rules#53.

PR #130, opened 2026-08-24
test(rules): mirror and cover PYD-008

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#54**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Pydantic AI was the last Python pack with tool discovery and no error-contract rule (CSDK-005, OAI-008, ADK-005, MCP-006 all exist). PYD-008 is framed around the split Pydantic AI actually makes: `ModelRetry` is appended to the message history and re-prompts the model within its `retries` budget, while any other exception propagates out of `agent.run()` and ends the run.

## What this PR does

1. Mirrors `pydantic_ai/error_handling.yaml` into `testdata/rules-fixture/`.

**Three cases rather than two**, because PYD-008 has two distinct silencing routes and both are worth pinning:

| case | expectation |
|---|---|
| `raise ValueError(...)`, uncaught | fires |
| try/except returning `{"error": ..., "retryable": ...}` | silent |
| `raise ModelRetry(...)` | silent |

The third is the one that matters most. The rule carries a `not: has_body_text: [ModelRetry]` clause purely as a **false-positive suppressor** — a tool already using the SDK's recovery channel is correct code and must not be reported. That case is what guards the suppressor against regression; without it, someone simplifying the rule to plain `has_raise` + `has_try_except` would break correct Pydantic AI code and the suite would stay green.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.194s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,36 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── PYD-008: Pydantic AI tool raises with no ModelRetry path ───────────
+	// has_raise + no try/except, minus a has_body_text suppressor for
+	// ModelRetry. Three cases: the raise, and BOTH silencing routes — catching
+	// the failure, and raising ModelRetry (the SDK's own recovery channel,
+	// which must not be reported as a defect).
+	{name: "PYD-008 fires on uncaught raise", ruleID: "PYD-008", kind: models.KindPydanticAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    if not order_id:
+        raise ValueError("order_id is required")
+    return order_id
+`, wantFires: true},
+	{name: "PYD-008 silent when the failure is caught", ruleID: "PYD-008", kind: models.KindPydanticAITool, src: `
+def lookup_order(order_id: str) -> dict:
+    """Look up an order."""
+    try:
+        if not order_id:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the engine half done the way the two-repo contract asks for: the fixture copy of `pydantic_ai/error_handling.yaml` is byte-identical to the pack in trustabl/trustabl-rules#54, so `rules-sync` will pass once both land. The test coverage is better than the minimum. Walking each case against the match expression: the uncaught `raise ValueError` fires, the try/except case is silenced by `has_try_except`, and the `raise ModelRetry` case is silenced by the `not: has_body_text` clause. That third case is the one I care about most, since it pins the suppressor: anyone later simplifying the rule to plain `has_raise` plus `has_try_except` would start flagging correct Pydantic AI code, and this case turns that into a test failure instead of a silent regression. The kind override to `KindPydanticAITool` matches how the harness forces kinds, so the bare functions parse fine. One trivial observation, the diff also removes a stray blank line in `policyRepoRuleCases`; harmless, no need to change anything. Approving, with the note that this should merge together with trustabl/trustabl-rules#54 (neither half alone), and that we will sort out the rulebook rationale doc for `PYD-008` on our side before the rule ships.

sairenchristianbuerano (approved):

PR #131, opened 2026-08-24
test(rules): mirror and cover CREW-009

bradAGI:
> Engine half of a coordinated pair. Rules half: **https://github.com/trustabl/trustabl-rules/pull/99**, on a branch of the **same name**, so `rules-sync` resolves the matching production pack rather than `main`. Neither half should merge alone.

## What the pair adds

Every pack except CrewAI ships a no-timeout rule — CSDK-003, OAI-005, ADK-003, MCP-004, AG2-012, PYD-006, VAI-011. The rules PR adds CREW-009 using the existing `call_without_kwarg` predicate.

What makes it matter in CrewAI specifically: `max_iter` bounds steps, not time, and in a sequential process a stalled tool call blocks every task queued behind it — one unresponsive endpoint stops the whole crew, not just the calling agent.

## What this PR does

Mirrors `crewai/network.yaml` into `testdata/rules-fixture/` and adds a fire and a silent case to `policyRuleCases`. The silent case is the same request with `timeout=10`, so it demonstrates the rule's own fix clears the finding.

## Verification

```

$ go test ./internal/rules/
```

End to end, a build of this branch against the paired rules branch:

```
CREW-009 [high] fetch_quote tools.py:6
```

with the `timeout=10` twin in the same file not firing.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -425,6 +425,19 @@ def fetch(q: str) -> str:
     return requests.get("https://api.example.com/data").text
 `, wantFires: false},
 
+	{name: "CREW-009 fires on request without timeout", ruleID: "CREW-009", kind: models.KindCrewAITool, src: `
+def fetch_quote(symbol: str) -> str:
+    """Fetch a price quote."""
+    import requests
+    return requests.get("https://quotes.example.com/" + symbol).text
+`, wantFires: true},
+	{name: "CREW-009 silent with timeout", ruleID: "CREW-009", kind: models.KindCrewAITool, src: `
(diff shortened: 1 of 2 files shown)


bradAGI:
Closing as a duplicate.

The no-timeout gap this fills is already covered by the LC-007/LC-015 and CREW-007 network PRs, which reached the same conclusion with different IDs. Keeping both would put two rules on one ID and break the loader.

The LangChain- and CrewAI-specific arguments are preserved in the commit here — that max_iterations/max_iter bounds steps and not time, and that in a sequential CrewAI process a stalled call blocks every task queued behind it.

PR #132, opened 2026-08-24
test(rules): mirror and cover LC-008

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#55**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

LangChain had no error-contract rule (CSDK-005, OAI-008, ADK-005, MCP-006 all exist). LC-008 is framed for LangChain's actual behavior: with the default `handle_tool_error=False` the exception propagates out of the `AgentExecutor` and aborts the run, and opting in with `handle_tool_error=True` only hands the model the raw `str(exception)`.

## What this PR does

1. Mirrors `langchain/error_handling.yaml` into `testdata/rules-fixture/`.

The silent case returns the `{"error": ..., "retryable": ...}` shape the rule's `fix` text prescribes rather than deleting the `raise`, so it demonstrates the prescribed remediation clears the finding.

## A gap this turned up

**Python only, deliberately.** I drafted a LangChain.js counterpart (LC-016) and dropped it after testing: `PredHasRaise` matches the Python grammar node `raise_statement`, so `has_raise` is structurally always false for a TypeScript tool — whose throws parse as `throw_statement`. The rule validated cleanly and never fired on a fixture that throws.

That blocks TS error-contract rules in *every* pack, not just this one. `PredHasShellCall` already branches on `models.IsTSOrJS` and reads a discovery-computed fact, so `PredHasRaise`/`PredHasTryExcept` could do the same. Happy to open that PR if you'd want it — it'd unblock a TS half for CSDK-005, OAI-008, ADK-005, MCP-006 and the four newer packs at once.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.149s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,25 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── LC-008: LangChain tool raises with no structured error contract ────
+	{name: "LC-008 fires on uncaught raise", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    if not order_id:
+        raise ValueError("order_id is required")
+    return order_id
+`, wantFires: true},
+	{name: "LC-008 silent when the failure is caught", ruleID: "LC-008", kind: models.KindLangChainTool, src: `
+def lookup_order(order_id: str) -> dict:
+    """Look up an order."""
+    try:
+        if not order_id:
+            raise ValueError("order_id is required")
+        return {"order_id": order_id}
+    except ValueError as exc:
+        return {"error": str(exc), "retryable": False}
+`, wantFires: false},
 }
 
 // policyRepoRuleCases covers repo-scoped rules.
@@ -2246,7 +2265,6 @@ var policyRepoRuleCases = []policyRepoCase{
 		},
 		models.RepoInventory{SDKsDetected: []models.SDK{models.SDKOpenAIAgents}},
 		false},
-
 }
 
 // optionsWithPermissionMode builds a ClaudeAgentOptionsDef whose captured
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done right. The fixture file is byte-identical to the pack in trustabl/trustabl-rules#55, and both test cases walk correctly against the match expression: the fire case has an uncaught `raise` so `has_raise` is true and `has_try_except` is false, and the silent case keeps a `raise` inside a try/except and returns the structured `{"error": ..., "retryable": ...}` shape. That last part is a nice touch, since it proves the prescribed fix clears the finding rather than just deleting the raise. Your `PredHasRaise` analysis also checks out on my read of `internal/rules/predicates.go`: it matches only the Python `raise_statement` node, so a TS counterpart would never fire. I would take a follow-up PR that makes `has_raise` and `has_try_except` language-aware along the lines of `PredHasShellCall`. One tiny nit: the diff also removes an unrelated blank line in `policyRepoRuleCases`, harmless but out of scope. Coordination note: `LC-008` is also claimed by trustabl/trustabl#122, trustabl/trustabl#123, trustabl/trustabl-rules#65, and trustabl/trustabl-rules#77, so whichever pair lands first, the others will need to renumber and rebase. On its own merits this is mergeable together with its rules half.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-008`.

Closing trustabl/agent-reliability-rules#65 (same rule, different author) and trustabl/agent-reliability-rules#77 (used `LC-008` for the idempotency rule instead).

sairenchristianbuerano (approved):

sairenchristianbuerano (approved):

PR #133, opened 2026-08-24
test(rules): mirror and cover CREW-008

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#56**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

CrewAI had no error-contract rule (CSDK-005, OAI-008, ADK-005, MCP-006 all exist). CREW-008 is framed for how CrewAI behaves: an uncaught exception is stringified into the agent's observation, so the agent typically burns its remaining `max_iter` budget re-calling the same tool the same way — and in a crew, that unusable output is threaded forward as context to every task behind it.

## What this PR does

1. Mirrors `crewai/error_handling.yaml` into `testdata/rules-fixture/`.

The silent case returns the **actionable error string** CREW-008's `fix` text prescribes, not the `{"error": ..., "retryable": ...}` dict the other packs' rules ask for. CrewAI hands the agent a string observation rather than a structured payload, so the remediation shape genuinely differs — the case reflects the SDK instead of copying the Claude SDK shape.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.165s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,25 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── CREW-008: CrewAI tool raises with no structured error contract ─────
+	{name: "CREW-008 fires on uncaught raise", ruleID: "CREW-008", kind: models.KindCrewAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    if not order_id:
+        raise ValueError("order_id is required")
+    return order_id
+`, wantFires: true},
+	{name: "CREW-008 silent when the failure is caught", ruleID: "CREW-008", kind: models.KindCrewAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    try:
+        if not order_id:
+            raise ValueError("order_id is required")
+        return order_id
+    except ValueError:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done exactly the way the two-repo contract asks. The fixture copy of `crewai/error_handling.yaml` is byte identical to the pack in trustabl/trustabl-rules#56, and the two `policyRuleCases` entries are genuinely non-vacuous: the silent case still contains a `raise`, so it exercises the `has_try_except` leg instead of trivially failing `has_raise`, and it returns the actionable error string the fix text prescribes rather than copying the structured-dict shape from the other SDK packs. That is the kind of test I want to see. Severity and confidence (low, 0.6) match CSDK-005 and OAI-008, which is the right calibration.

One coordination note on the branch-name trick described in the PR body: the `rules-sync` job resolves a same-named branch by asking `trustabl/trustabl-rules` itself for it, so a branch living on a fork will not be found and the job falls back to main. Expect this PR's rules-sync check to stay red until trustabl/trustabl-rules#56 merges; we will land the rules half first and this one immediately after. Nothing for you to do there. Approving.

PR #134, opened 2026-08-24
test(rules): mirror and cover AG2-013

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#57**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

AutoGen had no error-contract rule (CSDK-005, OAI-008, ADK-005, MCP-006 all exist). AG2-013 is framed for AutoGen's two-agent reply loop: tools execute on the executor's side and the result is sent back as that agent's reply, so an uncaught exception either aborts `initiate_chat` or lands in the message history as a stringified traceback that both agents keep reasoning over until AG2-004's `max_round` or AG2-006's auto-reply cap ends the chat.

## What this PR does

1. Mirrors `autogen/error_handling.yaml` into `testdata/rules-fixture/`.

The silent case returns the **actionable string** AG2-013's `fix` text prescribes rather than the `{"error": ..., "retryable": ...}` dict the Claude SDK and MCP rules ask for. AutoGen posts the tool result back as the executor agent's reply, so what the assistant reasons over is a message — the case matches the SDK rather than copying another pack's remediation shape.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	2.996s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,25 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── AG2-013: AutoGen tool raises with no structured error contract ─────
+	{name: "AG2-013 fires on uncaught raise", ruleID: "AG2-013", kind: models.KindAutoGenTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    if not order_id:
+        raise ValueError("order_id is required")
+    return order_id
+`, wantFires: true},
+	{name: "AG2-013 silent when the failure is caught", ruleID: "AG2-013", kind: models.KindAutoGenTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    try:
+        if not order_id:
+            raise ValueError("order_id is required")
+        return order_id
+    except ValueError:
+        return "error: order_id was empty; supply the customer's order number"
+`, wantFires: false},
 }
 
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is how a coordinated pair should land. The fixture copy is byte identical to the pack in trustabl/trustabl-rules#57, the cases use `models.KindAutoGenTool` like the rest of the AG2 coverage, and the silent case is the right kind of non-vacuous: it keeps the `raise` so it exercises the `has_try_except` leg of the match instead of trivially failing `has_raise`. Returning the actionable string the fix text prescribes, rather than the Claude SDK dict shape, carries the per-SDK framing into the tests too, which I appreciate.

Two small nits, neither blocking. The new cases are appended at the tail of `policyRuleCases` while every other AG2 case lives in the AutoGen block around line 450 of `internal/rules/policies_test.go`; grouping them there keeps the file navigable. And the hunk that drops the blank line at the end of `policyRepoRuleCases` is unrelated whitespace, fine either way.

On sequencing: the `rules-sync` job resolves a same-name branch only on the upstream trustabl-rules repo, so with the rules half coming from a fork this PR's sync check will stay red against production main until that half merges. That is expected for a pair. I will land trustabl/trustabl-rules#57 first and this immediately after so main never sits divergent.

PR #135, opened 2026-08-24
test(rules): add MCP-027, MCP-028, MCP-029 fire/silent cases

Paramjeet-singh-neu:
## Summary
- Mirror **MCP-027 / MCP-028 / MCP-029** into `testdata/rules-fixture/mcp/` so the engine fixture stays in sync with the rules pack.
- Add fire/silent `TestPolicyRules` cases:
  - MCP-027: `createCharge` without `idempotencyKey` fires; with key silent; `getBalance` silent
  - MCP-028: `registerTool("process", ...)` fires; `summarizeInvoice` silent
  - MCP-029: PHP `create_order` without `idempotency_key` fires; with key silent; `fetch_weather` silent

Paired with trustabl-rules and trustabl-rulebook on the same branch `feat/mcp-ts-php-parity`.

## Test plan
- [x] `go test ./internal/rules/ -run TestPolicyRules`
- [x] `RULES_REPO=../trustabl-rules scripts/check-rules-sync.sh`

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1389,6 +1389,56 @@ def run_cmd(name: str) -> str:
 		src: "use rmcp::tool;\nimpl T {\n    #[tool(description = \"Summarize\")]\n    fn summarize_invoice(&self) -> String { String::new() }\n}\n",
 	},
 	{
+		name: "MCP-027 fires on camelCase mutating tool with no idempotency param", ruleID: "MCP-027",
+		kind: models.KindMCPTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { McpServer } from \"@modelcontextprotocol/sdk/server/mcp.js\";\n" +
+			"const server = new McpServer({ name: \"s\", version: \"1.0.0\" });\n" +
+			"server.registerTool(\"createCharge\", { description: \"Charge a card\", inputSchema: { amount: z.number() } }, async ({ amount }) => ({ content: [{ type: \"text\", text: String(amount) }] }));\n",
+	},
+	{
+		name: "MCP-027 silent when idempotency key present", ruleID: "MCP-027",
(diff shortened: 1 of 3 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu. This is exactly how the fixture half of a rule PR should look. Both YAML files are byte-identical to the production pack in trustabl/trustabl-rules#100, so the rules-sync check will be happy, and every new rule has both a fire and a silent case. I walked the cases against discovery and they hold up: the TypeScript cases work because Zod `inputSchema` keys are captured as `ParamNames`, so `idempotencyKey` satisfies the case-insensitive `contains: idempot` match in the silent case, and the PHP cases work because the `#[McpTool]` attribute supplies the tool name while the method signature supplies the params, so `$idempotency_key` correctly silences MCP-029. The `getBalance` and `fetch_weather` negative cases exercise the prefix gate rather than just the param gate, which is a nice touch.

The only thing holding this pair up is coordination, not quality: MCP-027 and MCP-028 are also claimed by trustabl/trustabl#168 and trustabl/trustabl-rules#91, and MCP-029 by trustabl/trustabl#124, trustabl/trustabl#172, trustabl/trustabl-rules#80, and trustabl/trustabl-rules#94. Whichever set lands first keeps the IDs, and the rest will need to renumber and rebase, so this may need a mechanical renumber before merge. I will keep it in lockstep with the rules-repo PR either way.

sairenchristianbuerano:
**Closing.** All three ids resolve elsewhere:

- `MCP-027`, `MCP-028` → trustabl/agent-reliability-rules#91 (TypeScript description quality)
- `MCP-029` → trustabl/agent-reliability-rules#94 (TypeScript filesystem write)

Your three rules — TypeScript MCP idempotency, ambiguous TypeScript MCP tool name, and PHP MCP idempotency — are all distinct from those and are not shipping as a result. `MCP-030` onward is free if anyone wants to bring them back.

PR #136, opened 2026-08-24
test(rules): mirror and cover PYD-009

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#58**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Pydantic AI had no path-safety rule (CSDK-004, OAI-006, ADK-004, MCP-005 all exist). The SDK-specific point is that Pydantic AI's type layer looks like it covers this and doesn't: a parameter annotated `str` or `Path` validates cleanly while still carrying `../../etc/passwd`, because the annotation constrains the value's *type*, not the region of the filesystem it points at.

## What this PR does

1. Mirrors `pydantic_ai/path_safety.yaml` into `testdata/rules-fixture/`.

Three cases, following CSDK-004's own table:

| case | expectation |
|---|---|

The third guards the per-param behavior of `call_uses_unnormalized_path_param` — it's what keeps the rule from broadening into "any tool with a string param" if the predicate is ever reworked.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.112s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,29 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── PYD-009: Pydantic AI path param reaching I/O unnormalized ──────────
+	// call_uses_unnormalized_path_param is per-param, so the third case pins
+	// that a non-pathish param does not drag the rule in.
+	{name: "PYD-009 fires on path param in open()", ruleID: "PYD-009", kind: models.KindPydanticAITool, src: `
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    with open(note_path, "r") as f:
+        return f.read()
+`, wantFires: true},
+	{name: "PYD-009 silent with .resolve()", ruleID: "PYD-009", kind: models.KindPydanticAITool, src: `
+from pathlib import Path
+def read_note(note_path: str) -> str:
+    """Read a note."""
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is exactly how I want the engine half of a rule pair to land. I verified the fixture file is byte identical to the pack in trustabl/trustabl-rules#58 (same git blob), so `rules-sync` stays green as long as the two merge together. The rule also checks out against main: `call_uses_unnormalized_path_param` exists with exactly the `callees` and `callee_prefixes` shape you use, `pydantic_ai_tool` is a valid tool scope token, and no `schema_version` bump is needed since there is no new predicate. I walked all three cases against the predicate as written: `open(note_path)` fires, `Path(note_path).resolve()` puts the param in the normalized set so the second case stays silent, and `note_id` never enters the pathish set, which pins the per-param gate the way your comment describes. Good call on the ID too, `PYD-008` is already claimed by the open pair trustabl/trustabl#130 and trustabl/trustabl-rules#54, so `PYD-009` avoids a collision. One trivial nit: the diff also drops a blank line inside `policyRepoRuleCases`, an unrelated whitespace change I would normally leave out, but it is not worth a rebase on its own. Approving, to merge in lockstep with the rules half.

PR #137, opened 2026-08-24
test(rules): mirror and cover LC-009, LC-023

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#59**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

LangChain had no path-safety rule (CSDK-004/012, OAI-006, ADK-004, MCP-005 all exist). The rules PR ships the python/TypeScript pair this pack uses throughout: LC-009 (per-param `call_uses_unnormalized_path_param`, same callee set as CSDK-004) and LC-023 (coarse `has_write_call`, mirroring CSDK-012 until TS path-normalization analysis exists).

## What this PR does

1. Mirrors `langchain/path_safety.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| LC-009 — `open(note_path)` on a raw param | fires |
| LC-009 — `Path(note_path).resolve()` first | silent |
| LC-009 — non-pathish param (`note_id`), no I/O | silent |
| LC-023 — `writeFileSync(p, body)` | fires |
| LC-023 — in-memory store, no filesystem write | silent |

LC-009's third case guards the per-param behavior of `call_uses_unnormalized_path_param` — it's what keeps the rule from broadening into "any tool with a string param" if the predicate is ever reworked.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.210s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,50 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── LC-009 / LC-023: LangChain path safety (python + typescript) ───────
+	// LC-009 is the per-param Python flow check; LC-023 is the coarse TS
+	// filesystem-write signal, mirroring CSDK-012 until TS path normalization
+	// analysis exists.
+	{name: "LC-009 fires on path param in open()", ruleID: "LC-009", kind: models.KindLangChainTool, src: `
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    with open(note_path, "r") as f:
+        return f.read()
+`, wantFires: true},
+	{name: "LC-009 silent with .resolve()", ruleID: "LC-009", kind: models.KindLangChainTool, src: `
+from pathlib import Path
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    p = Path(note_path).resolve()
+    with open(p, "r") as f:
+        return f.read()
+`, wantFires: false},
+	{name: "LC-009 silent on non-pathish param", ruleID: "LC-009", kind: models.KindLangChainTool, src: `
+def get_note_meta(note_id: str) -> dict:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done right. The fixture file is byte-identical to `langchain/path_safety.yaml` in trustabl/trustabl-rules#59 (same blob), so the sync check is genuinely satisfied rather than approximately. The test cases are the part I looked at hardest and they hold up. The LC-009 trio mirrors the established CSDK-004 patterns closely: the `.resolve()` silent case is exactly the shape the predicate's normalized-param tracking already handles, and the non-pathish `note_id` case pins the pathish-name gating, which is the per-param guard your PR body describes. The LC-023 TS cases import from `@langchain/core/tools`, which the harness needs for discovery to produce the tool at all, and bare `writeFileSync` is in the TS handler-facts write set, so the fire case exercises the real path rather than passing vacuously. Reading the match expressions against the snippets, every fire and silent expectation checks out.

Coordination note: `LC-009` is also claimed by trustabl/trustabl#122 and trustabl/trustabl-rules#65, so whichever pair merges first, the others renumber and rebase. And as your description already says, this must land together with trustabl/trustabl-rules#59 or rules-sync goes red; I have left the same note there. No blocking issues from me.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-009` and `LC-023`.

Closing trustabl/agent-reliability-rules#65 + trustabl/agent-reliability-analyzer#122, which used `LC-009` for a LangChain idempotency rule. That rule ships as `LC-017` via trustabl/agent-reliability-rules#62, so `LC-009` stays with path safety here.

sairenchristianbuerano (approved):

PR #138, opened 2026-08-24
test(rules): mirror and cover AG2-015

bradAGI:
> Engine half of a coordinated pair. Rules half: **https://github.com/trustabl/trustabl-rules/pull/101**, on a branch of the **same name**, so `rules-sync` resolves the matching production pack rather than `main`. Neither half should merge alone.

## What the pair adds

Six packs ship an idempotency rule — CSDK-006/016, OAI-009/019, ADK-006, MCP-007, CREW-006, PYD-007. AutoGen ships none. The rules PR adds AG2-015 using the existing `name_has_prefix` + `param_name_matches` predicate pair.

What makes it matter in AutoGen specifically: the tool response is a message in a conversation re-sent in full on every later turn, so an inconclusive call stays re-invitable for the whole run — and in a group chat any speaker the manager selects can re-issue it, including an agent that never saw the first attempt.

## What this PR does

Mirrors `autogen/idempotency.yaml` into `testdata/rules-fixture/` and adds a fire and a silent case to `policyRuleCases`.

The silent case is the same `refund_payment` with an `idempotency_key` parameter **threaded through to the backing call**, so it demonstrates the rule's own fix clears the finding — rather than that renaming the tool out of the mutating-prefix set does.

## Verification

```

$ go test ./internal/rules/
```

End to end, a build of this branch against the paired rules branch:

```
AG2-015 [medium] refund_payment tools.py:6
```

with the keyed twin in the same file not firing.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -510,6 +510,17 @@ def fetch(q: str) -> str:
     return requests.get("https://api.example.com/data").text
 `, wantFires: false},
 
+	{name: "AG2-015 fires on mutating tool without key", ruleID: "AG2-015", kind: models.KindAutoGenTool, src: `
+def refund_payment(payment_id: str, amount: int) -> str:
+    """Refund a payment."""
+    return billing.refund(payment_id, amount)
+`, wantFires: true},
+	{name: "AG2-015 silent with idempotency key", ruleID: "AG2-015", kind: models.KindAutoGenTool, src: `
+def refund_payment(payment_id: str, amount: int, idempotency_key: str) -> str:
+    """Refund a payment."""
+    return billing.refund(payment_id, amount, idempotency_key=idempotency_key)
+`, wantFires: false},
+
 	{name: "AG2-012 fires on network call without timeout", ruleID: "AG2-012", kind: models.KindAutoGenTool, src: `
 def fetch(q: str) -> str:
     """Fetch."""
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture file here is byte identical to the pack in trustabl/trustabl-rules#101 (same blob), so the two-repo sync contract is satisfied once both halves land together. The test cases are exactly what I want to see: the fire case (`refund_payment` with no key) walks correctly through `name_has_prefix` plus the negated `param_name_matches`, and the silent case demonstrates the rule's own fix, an `idempotency_key` parameter threaded through to the backing call, rather than being silenced by renaming the tool out of the mutating prefix set. That distinction matters for test quality and you clearly thought about it. Both predicates exist in the engine on main and `autogen_tool` is a valid tool scope kind, so no schema or manifest change is needed, and the PR correctly makes none.

Same coordination note as on the rules half: trustabl/trustabl#142 paired with trustabl/trustabl-rules#63 is your earlier open pair claiming the same `AG2-015`, and one of the two pairs should be closed before either merges. This branch looks like the one to keep, but just confirm which. Once the rule lands we also owe it a rationale doc in trustabl-rulebook, which the maintainers can take care of or pair with you on.

sairenchristianbuerano (approved):

sairenchristianbuerano:
**Selected for merge.** This pair keeps `AG2-015`.

Closing trustabl/agent-reliability-rules#63 + trustabl/agent-reliability-analyzer#142 — the same rule submitted twice by the same author on a second branch. Identical title, severity, confidence and match expression; this pair is the one without merge conflicts.

PR #139, opened 2026-08-24
test(rules): mirror and cover CREW-009

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#60**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

CrewAI had no path-safety rule (CSDK-004, OAI-006, ADK-004, MCP-005 all exist). Two CrewAI-specific points shape the rule text: the `args_schema` constrains the argument's *type*, not where it points, so a traversal payload validates cleanly; and the reach is wider than the owning agent, since a crew shares one process, threads task output forward as context, and CREW-104 delegation lets a peer reach the tool.

## What this PR does

1. Mirrors `crewai/path_safety.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

The third guards the per-param behavior of `call_uses_unnormalized_path_param` — it's what keeps the rule from broadening into "any tool with a string param" if the predicate is ever reworked.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── CREW-009: CrewAI path param reaching I/O unnormalized ──────────────
+	{name: "CREW-009 fires on path param in open()", ruleID: "CREW-009", kind: models.KindCrewAITool, src: `
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    with open(note_path, "r") as f:
+        return f.read()
+`, wantFires: true},
+	{name: "CREW-009 silent with .resolve()", ruleID: "CREW-009", kind: models.KindCrewAITool, src: `
+from pathlib import Path
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    p = Path(note_path).resolve()
+    with open(p, "r") as f:
+        return f.read()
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done right. The fixture file is byte-identical to the production YAML in trustabl/trustabl-rules#60, so `rules-sync` is satisfied once the two land together, and I appreciate the PR body spelling out that coordination and the same-name branch trick. Walking the three cases against the predicate, none of them pass vacuously: `open(note_path)` on a raw pathish param fires, `Path(note_path).resolve()` marks `note_path` normalized through the resolve-receiver walk so the second stays silent, and `note_id` fails the pathish check entirely, so the third genuinely guards the per-param gating rather than restating the second case. That last one is a thoughtful addition.

Two small nits, neither blocking. The new cases sit at the tail of `policyRuleCases` instead of under the existing CrewAI tool rules section in `internal/rules/policies_test.go`, where the other CREW cases live, and the second hunk deletes an unrelated blank line in `policyRepoRuleCases`. I would fold the cases into the CREW section for tidiness, but I am happy to take this as is. Separately, before the rule ships we need a rationale doc in the trustabl-rulebook repo; the maintainers can handle that or pair on it, it is not a demand on you. Approving alongside the rules half.

PR #140, opened 2026-08-24
test(rules): mirror and cover AG2-014

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#61**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

AutoGen had no path-safety rule (CSDK-004, OAI-006, ADK-004, MCP-005 all exist). The AutoGen-specific point in the rule text is one people reasonably assume they're already covered on: **a registered tool sidesteps the containment AG2-001 is about.** Generated code may run in a Docker executor, but a function registered with `register_for_execution` runs in the host process with the host's filesystem view regardless — so "we run code in Docker" isn't an answer to this finding, and the check has to live inside the registered function.

## What this PR does

1. Mirrors `autogen/path_safety.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

The third guards the per-param behavior of `call_uses_unnormalized_path_param`.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── AG2-014: AutoGen path param reaching I/O unnormalized ──────────────
+	{name: "AG2-014 fires on path param in open()", ruleID: "AG2-014", kind: models.KindAutoGenTool, src: `
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    with open(note_path, "r") as f:
+        return f.read()
+`, wantFires: true},
+	{name: "AG2-014 silent with .resolve()", ruleID: "AG2-014", kind: models.KindAutoGenTool, src: `
+from pathlib import Path
+def read_note(note_path: str) -> str:
+    """Read a note."""
+    p = Path(note_path).resolve()
+    with open(p, "r") as f:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the engine half done exactly the way the two-repo contract asks: the fixture copy of `autogen/path_safety.yaml` is byte identical to the pack in trustabl/trustabl-rules#61 (same blob in both diffs), and the three cases are the right three. Walking them against the predicate on main: `note_path` is pathish and reaches `open` unresolved, so the fire case fires; `Path(note_path).resolve()` marks the param normalized before the `Path` callee check can trip, so the resolve case stays silent; and `note_id` never enters the pathish set, so the third case exercises the gate you intended. The comment above the `match:` block defending the per-param choice mirrors CSDK-004's and is worth keeping. Two small things. The hunk removing the blank line at the end of `policyRepoRuleCases` is unrelated to this change; it is harmless, but I would drop it to keep the diff clean. And the one-word wording fix I asked for on the rules side ("crew" is CrewAI vocabulary, not AutoGen's) needs mirroring into the fixture copy here so `rules-sync` stays green. With that mirrored, this is good to go, and I appreciate that the PR body spells out the pairing so neither half merges alone.

PR #141, opened 2026-08-24
test(rules): mirror and cover LC-017

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#62**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

LangChain had no idempotency rule (CREW-006, PYD-007, MCP-007, CSDK-006, OAI-009 all exist). The LangChain-specific argument is that **repetition is routine here, not exceptional**: an `AgentExecutor` step whose observation reads as inconclusive — a timeout, a truncated response, an error string returned under `handle_tool_error` — leads the model to re-issue the same action, with no way to know the first call already committed. `with_retry` compounds it, since a network failure *after* the side effect landed is indistinguishable from one before.

LC-017 uses the same `name_has_prefix` + `param_name_matches` shape as MCP-007.

## What this PR does

1. Mirrors `langchain/idempotency.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| `refund_payment(charge_id, amount_cents)` | fires |
| same tool with an `idempotency_key` param | silent |
| `get_payment(charge_id)` — read-only name | silent |

The third case pins **both halves** of the match: a missing key alone isn't enough, the name has to signal a side effect. Without it the rule could quietly broaden into "every tool without an idempotency key" and the suite would stay green.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,25 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── LC-017: mutating LangChain tool with no idempotency key ────────────
+	// Same name_has_prefix + param_name_matches shape as MCP-007. Third case
+	// pins that a non-mutating name is not swept in by the missing key alone.
+	{name: "LC-017 fires on mutating tool without key", ruleID: "LC-017", kind: models.KindLangChainTool, src: `
+def refund_payment(charge_id: str, amount_cents: int) -> dict:
+    """Refund a charge."""
+    return {"ok": True}
+`, wantFires: true},
+	{name: "LC-017 silent with idempotency key", ruleID: "LC-017", kind: models.KindLangChainTool, src: `
+def refund_payment(charge_id: str, amount_cents: int, idempotency_key: str) -> dict:
+    """Refund a charge."""
+    return {"ok": True}
+`, wantFires: false},
+	{name: "LC-017 silent on a read-only tool name", ruleID: "LC-017", kind: models.KindLangChainTool, src: `
+def get_payment(charge_id: str) -> dict:
+    """Get a charge."""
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the engine half done right: the fixture copy of `langchain/idempotency.yaml` is byte-identical to the file in trustabl/trustabl-rules#62 (same blob hash), so `rules-sync` will pass once the two merge together, and the coordinated-branch note in the PR body is exactly the workflow we want.

I walked the three cases against the match expression and they all hold. `refund_payment(charge_id, amount_cents)` hits the `refund_` prefix with no parameter matching `idempot`, `request_id`, or `txn_id`, so it fires. Adding `idempotency_key` satisfies the `contains: idempot` arm and goes silent. And the `get_payment` case is the one I am happiest to see, since it pins the `name_has_prefix` half of the conjunction; without it the rule could quietly degrade into flagging every keyless tool and the suite would stay green. The cases follow the same `models.KindLangChainTool` shape as the existing LC Python cases, so they slot into the harness cleanly.

One tiny incidental: the diff also drops a blank line at the end of `policyRepoRuleCases`. Harmless, no need to change it. This is good to go from my side; I will land it in lockstep with the rules half, and we will add the rulebook rationale doc for `LC-017` on our end.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-017`.

The LangChain idempotency rule was submitted three times under three ids — `LC-008` (trustabl/agent-reliability-rules#77), `LC-009` (trustabl/agent-reliability-rules#65) and `LC-017` here. `LC-017` is the only one that collides with nothing, so it wins and the other two close.

PR #142, opened 2026-08-24
test(rules): mirror and cover AG2-015

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#63**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

AutoGen had no idempotency rule (CREW-006, PYD-007, MCP-007, CSDK-006, OAI-009 all exist). AutoGen's conversation loop makes repetition **the default failure mode**: the assistant proposes, the executor runs and replies, and an inconclusive reply leads the assistant to propose the same call again next round with no way to know the first committed. AG2-004's `max_round` and AG2-006's auto-reply cap **bound how many duplicates land rather than preventing them** — and in a group chat a second agent reading the same transcript can re-propose independently.

## What this PR does

1. Mirrors `autogen/idempotency.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| `refund_payment(charge_id, amount_cents)` | fires |
| same tool with an `idempotency_key` param | silent |
| `get_payment(charge_id)` — read-only name | silent |

The third case pins **both halves** of the match: a missing key alone isn't enough, the name has to signal a side effect.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,23 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── AG2-015: mutating AutoGen tool with no idempotency key ─────────────
+	{name: "AG2-015 fires on mutating tool without key", ruleID: "AG2-015", kind: models.KindAutoGenTool, src: `
+def refund_payment(charge_id: str, amount_cents: int) -> dict:
+    """Refund a charge."""
+    return {"ok": True}
+`, wantFires: true},
+	{name: "AG2-015 silent with idempotency key", ruleID: "AG2-015", kind: models.KindAutoGenTool, src: `
+def refund_payment(charge_id: str, amount_cents: int, idempotency_key: str) -> dict:
+    """Refund a charge."""
+    return {"ok": True}
+`, wantFires: false},
+	{name: "AG2-015 silent on a read-only tool name", ruleID: "AG2-015", kind: models.KindAutoGenTool, src: `
+def get_payment(charge_id: str) -> dict:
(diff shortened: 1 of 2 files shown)


jhumel-code:
@bradAGI thanks, this is the model shape for the engine half of a rule contribution. The fixture copy of `autogen/idempotency.yaml` is identical down to the byte with trustabl/trustabl-rules#63, and the three cases pin both halves of the conjunction: one fire, one silent through the `idempotency_key` parameter, and one silent through a read-only name, so neither conjunct can regress without a test catching it. I walked each snippet against the match expression as written and they behave as labeled.

The one substantive issue is a duplicate on your own side: trustabl/trustabl#138, paired with trustabl/trustabl-rules#101, claims AG2-015 on a different branch with different explanation text. Both pairs cannot land, and whichever arrives second will fail loading on a duplicate ID, so please pick one and close the other (I lean toward this version). One tiny nit, the diff also deletes a blank line inside `policyRepoRuleCases` that has nothing to do with this change; it is harmless but will conflict with your other open PRs touching the same file, so I would drop it. Once the duplicate question is settled I am glad to take this together with the rules half.

sairenchristianbuerano:
**Closing in favour of trustabl/agent-reliability-rules#101 + trustabl/agent-reliability-analyzer#138**, which is the same `AG2-015` rule you submitted on a second branch — identical title, severity, confidence and match expression. That pair merges without conflicts, this one needs a rebase, so we kept the clean one. Nothing is lost.

PR #143, opened 2026-08-24
test(rules): mirror and cover PYD-010, PYD-011

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#64**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to Pydantic AI. PYD-001 only checks that a docstring *exists*, so `"""TODO: describe this tool."""` and `"""Gets data."""` pass today. The cost is higher in this pack than in the Claude SDK: Pydantic AI also parses the docstring's parameter sections into the argument schema, so a placeholder strips per-argument guidance at the same time as tool-level guidance.

## What this PR does

1. Mirrors `pydantic_ai/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| PYD-010 — `"""TODO: describe this tool."""` | fires |
| PYD-010 — real description | silent |
| PYD-011 — `"""Gets data."""` | fires |
| PYD-011 — full sentence | silent |
| PYD-011 — **no docstring at all** | silent |

**Five cases rather than four.** PYD-011 pairs `description_length_lt: 40` with `has_docstring: true` specifically so an absent docstring stays PYD-001's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case is what pins that guard; without it, someone dropping the `has_docstring` conjunct would make both rules fire on the same tool and the suite would stay green.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,35 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── PYD-010 / PYD-011: Pydantic AI description quality ─────────────────
+	// PYD-010 is has_description_text (placeholder markers); PYD-011 pairs
+	// description_length_lt with has_docstring so an ABSENT docstring stays
+	// PYD-001's finding rather than double-reporting here.
+	{name: "PYD-010 fires on placeholder description", ruleID: "PYD-010", kind: models.KindPydanticAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """TODO: describe this tool."""
+    return order_id
+`, wantFires: true},
+	{name: "PYD-010 silent on a real description", ruleID: "PYD-010", kind: models.KindPydanticAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up a single order by its identifier and return its current status."""
+    return order_id
+`, wantFires: false},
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done exactly the way the two-repo contract asks. The fixture copy of `pydantic_ai/tool_definition.yaml` is byte identical to the production file in trustabl/trustabl-rules#64 (both diffs carry the same blob hash), so the rules-sync check should be satisfied once the two land together. The test cases are the part worth calling out. Walking them against the predicates on main, both fire and silent cases behave as written (`has_description_text` lowers both sides, so the `TODO:` docstring matches the `todo` needle, and the silent `PYD-011` description clears the 40 character threshold), and the fifth case, `PYD-011` silent when the docstring is absent, is the kind of guard-pinning case I wish more rule PRs included. Without it, someone dropping the `has_docstring` conjunct would make `PYD-011` double-report against `PYD-001` while the suite stayed green, so that case carries real weight. One trivial nit: the diff also deletes a blank line inside `policyRepoRuleCases`, an unrelated whitespace touch, not worth a respin on its own. I will merge this together with the rules half, and we will take care of the rulebook rationale doc on our side before the rules ship. Approving.

PR #145, opened 2026-08-24
test(rules): mirror and cover LC-018, LC-019

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#66**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to LangChain. LC-001 only checks that a docstring *exists*, so `"""TODO: describe this tool."""` and `"""Gets data."""` pass today. Two things make the LangChain framing its own: mis-selection scales badly because agents are routinely handed a dozen or more tools at once, and each wrong pick burns an iteration against LC-102's `max_iterations` budget — so a run can exhaust its steps and return nothing useful.

## What this PR does

| case | expectation |
|---|---|
| LC-018 — `"""TODO: describe this tool."""` | fires |
| LC-018 — real description | silent |
| LC-019 — `"""Gets data."""` | fires |
| LC-019 — full sentence | silent |
| LC-019 — **no docstring at all** | silent |

**Five cases rather than four.** LC-019 pairs `description_length_lt: 40` with `has_docstring: true` so an absent docstring stays LC-001's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case pins that guard.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,32 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── LC-018 / LC-019: LangChain description quality ─────────────────────
+	{name: "LC-018 fires on placeholder description", ruleID: "LC-018", kind: models.KindLangChainTool, src: `
+def lookup_order(order_id: str) -> str:
+    """TODO: describe this tool."""
+    return order_id
+`, wantFires: true},
+	{name: "LC-018 silent on a real description", ruleID: "LC-018", kind: models.KindLangChainTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up a single order by its identifier and return its current status."""
+    return order_id
+`, wantFires: false},
+	{name: "LC-019 fires on a too-short description", ruleID: "LC-019", kind: models.KindLangChainTool, src: `
+def list_orders(customer_id: str) -> str:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. I compared the fixture copy against the rules half, trustabl/trustabl-rules#66, and the sixty added lines are identical, so the sync check has nothing to catch here as long as the two land together.

The cases hold up when I walk them against the match expressions. `parsePy` fills `Description` from the docstring, so "TODO: describe this tool." reaches `has_description_text` and fires `LC-018`, "Gets data." is ten characters and fires `LC-019`, and the silent description is sixty-two, comfortably past the threshold. Adding the fifth case for the absent docstring was the right instinct: that is exactly the guard that would rot quietly if someone later dropped `has_docstring: true` from `LC-019`, and a pinned case is the only thing that would catch it.

One nit, the diff also drops a blank line at the end of `policyRepoRuleCases`, unrelated to this change. Harmless, but the change reads cleaner without it.

I left a couple of non-blocking notes on the rules PR about the `todo` needle and the rule IDs quoted in the explanation text, so there is no need to answer them in both places. Neither half should merge on its own.

PR #146, opened 2026-08-24
test(rules): mirror and cover CREW-010, CREW-011

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#67**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to CrewAI. CREW-001 only checks that a docstring *exists*, so `"""TODO: describe this tool."""` and `"""Gets data."""` pass today. What makes the CrewAI framing its own is that **the consequence travels**: an agent that picks the wrong tool produces a task output threaded forward as context to every task behind it, and CREW-104 delegation lets peers reach the tool on the strength of the same placeholder.

## What this PR does

1. Mirrors `crewai/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| CREW-010 — `"""TODO: describe this tool."""` | fires |
| CREW-010 — real description | silent |
| CREW-011 — `"""Gets data."""` | fires |
| CREW-011 — full sentence | silent |
| CREW-011 — **no docstring at all** | silent |

**Five cases rather than four.** CREW-011 pairs `description_length_lt: 40` with `has_docstring: true` so an absent docstring stays CREW-001's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case pins that guard.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,32 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── CREW-010 / CREW-011: CrewAI description quality ────────────────────
+	{name: "CREW-010 fires on placeholder description", ruleID: "CREW-010", kind: models.KindCrewAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """TODO: describe this tool."""
+    return order_id
+`, wantFires: true},
+	{name: "CREW-010 silent on a real description", ruleID: "CREW-010", kind: models.KindCrewAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up a single order by its identifier and return its current status."""
+    return order_id
+`, wantFires: false},
+	{name: "CREW-011 fires on a too-short description", ruleID: "CREW-011", kind: models.KindCrewAITool, src: `
+def list_orders(customer_id: str) -> str:
+    """Gets data."""
+    return customer_id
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy is byte identical to the YAML in trustabl/trustabl-rules#67, so `rules-sync` will be satisfied as long as the two land together, and I plan to take them as a pair.

The cases hold up when walked against the predicates. `PredHasDescriptionText` lowercases both sides, so `"""TODO: describe this tool."""` matches the `todo` needle; both of the realistic descriptions are comfortably past the 40 character threshold (71 and 62), so the silent cases are not passing vacuously; and `PredHasDocstring` trims before comparing, so the absent-docstring case leaves `CREW-011` silent and lets `CREW-001` own that finding. Adding the fifth case rather than the minimum four was the right instinct, since it pins the exact guard that `has_docstring: true` exists for.

Two nits, neither blocking. The new cases are appended at the end of `policyRuleCases` rather than next to the existing `CREW-001` through `CREW-006` block, and they would be easier to find grouped with their neighbors. The diff also deletes an unrelated blank line at the end of `policyRepoRuleCases`; if you push again, please drop that hunk so the change stays scoped to the two rules.

PR #147, opened 2026-08-24
test(rules): mirror and cover AG2-016, AG2-017

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#68**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to AutoGen. AG2-007 only checks that a docstring *exists*, so `"""TODO: describe this tool."""` and `"""Gets data."""` pass today. **How AutoGen fails on a bad guess is what makes it expensive**: a wrongly chosen tool returns something that doesn't answer the assistant's intent, the assistant reads that as inconclusive and proposes another call, and the pair trades rounds until AG2-004's `max_round` or AG2-006's auto-reply cap ends the chat — with nothing in the transcript naming the stub description as the cause.

## What this PR does

1. Mirrors `autogen/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| AG2-016 — `"""TODO: describe this tool."""` | fires |
| AG2-016 — real description | silent |
| AG2-017 — `"""Gets data."""` | fires |
| AG2-017 — full sentence | silent |
| AG2-017 — **no docstring at all** | silent |

**Five cases rather than four.** AG2-017 pairs `description_length_lt: 40` with `has_docstring: true` so an absent docstring stays AG2-007's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case pins that guard.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,32 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── AG2-016 / AG2-017: AutoGen description quality ─────────────────────
+	{name: "AG2-016 fires on placeholder description", ruleID: "AG2-016", kind: models.KindAutoGenTool, src: `
+def lookup_order(order_id: str) -> str:
+    """TODO: describe this tool."""
+    return order_id
+`, wantFires: true},
+	{name: "AG2-016 silent on a real description", ruleID: "AG2-016", kind: models.KindAutoGenTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up a single order by its identifier and return its current status."""
+    return order_id
+`, wantFires: false},
+	{name: "AG2-017 fires on a too-short description", ruleID: "AG2-017", kind: models.KindAutoGenTool, src: `
+def list_orders(customer_id: str) -> str:
+    """Gets data."""
+    return customer_id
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the engine half done right. I checked that the fixture copy of `autogen/tool_definition.yaml` produces a byte-identical blob to the production file in trustabl/trustabl-rules#68, so `rules-sync` will be green once the two land together. The test cases are the correct shape (the same `parsePy` harness path the CSDK-017/018 cases use, where Python discovery populates `Description` from the docstring, so none of these pass vacuously): the AG2-016 fire case trips the case-insensitive `todo` needle, its silent case contains no needle, the AG2-017 fire case is 10 trimmed characters against the 40 threshold, and its silent case clears it comfortably. The fifth case pinning the no-docstring boundary against AG2-007 is a thoughtful touch, that guard is exactly the kind of thing that silently regresses otherwise. The stray blank-line removal in `policyRepoRuleCases` is unrelated to this change but harmless, I am fine leaving it. Neither ID collides with anything on main or with any other open PR. Approving, and I will merge this together with the rules half so neither side breaks the sync check alone.

PR #148, opened 2026-08-24
test(rules): mirror and cover VAI-013, VAI-014

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#69**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to the Vercel AI SDK. VAI-004 only checks that a `description` *exists*, so `"TODO: describe this tool."` and `"Gets data."` pass today — while leaving the tool in exactly the state VAI-004 exists to prevent. **VAI-004's own text says why this matters more here:** the SDK has no docstring fallback, so the description string is the entire prompt-side account of the tool.

## What this PR does

1. Mirrors `vercel_ai/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| VAI-013 — `description: "TODO: describe this tool."` | fires |
| VAI-013 — real description | silent |
| VAI-014 — `description: "Gets data."` | fires |
| VAI-014 — full sentence | silent |
| VAI-014 — **no `description` field at all** | silent |

**Five cases rather than four.** VAI-014 pairs `description_length_lt: 40` with `has_docstring: true` so an absent description stays VAI-004's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case pins that guard.

Both predicates read `ToolDef.Description`, which for TypeScript is the explicit `description` field — confirmed firing against TS tools rather than assumed.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,40 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── VAI-013 / VAI-014: Vercel AI description quality (typescript) ──────
+	// has_description_text and description_length_lt both read
+	// ToolDef.Description, which for TS is the explicit `description` field.
+	{
+		name: "VAI-013 fires on placeholder description", ruleID: "VAI-013",
+		kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { tool } from \"ai\";\n" +
+			"export const t = tool({ description: \"TODO: describe this tool.\", inputSchema: {}, execute: async () => \"x\" });\n",
+	},
+	{
+		name: "VAI-013 silent on a real description", ruleID: "VAI-013",
+		kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: false,
+		src: "import { tool } from \"ai\";\n" +
+			"export const t = tool({ description: \"Look up a single order by its identifier and return its status.\", inputSchema: {}, execute: async () => \"x\" });\n",
+	},
+	{
+		name: "VAI-014 fires on a too-short description", ruleID: "VAI-014",
+		kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: true,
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the engine half of trustabl/trustabl-rules#69 and it does exactly what the sync contract asks: the fixture copy of `vercel_ai/tool_definition.yaml` is byte identical to the production diff, and each new rule gets both a fire and a silent case in `policies_test.go`. I walked the five cases against the predicates as they exist on main and they all behave as labeled. The placeholder case hits the case-insensitive `todo` needle, the full-sentence description clears the 40-character bound, and the fifth case pins the guard I most wanted pinned: an absent description stays VAI-004's finding because `has_docstring` is false there, so VAI-014 stays silent. Going to five cases instead of four for that reason was the right call.

Two small notes. The diff also drops a stray blank line in `policyRepoRuleCases`, which is harmless but unrelated to this change. And VAI-013 is also claimed by trustabl/trustabl#121 and trustabl/trustabl-rules#102, so whichever pair lands first, the others will need to renumber and rebase; judged on its own, this PR is fine. A rulebook rationale doc is needed before the rules ship, which we can handle on our side or pair on. Mergeable together with its rules half once the ID collision shakes out.

sairenchristianbuerano:

Nothing needed from you.

sairenchristianbuerano:
For the record, since this PR was involved in an id clash: `VAI-013` had also been claimed by the privileged-tool approval rule in trustabl/agent-reliability-rules#102 / trustabl/agent-reliability-analyzer#121.

**This PR keeps `VAI-013` and `VAI-014`.** We renumbered the other rule to `VAI-018` instead, because this one claims the two ids as a contiguous pair and was opened first in the rules repo. Nothing changed here as a result — no action needed.

Verified: the validator reports 220 rules valid with both pairs applied, no duplicate ids.

sairenchristianbuerano (approved):

PR #150, opened 2026-08-24
test(rules): mirror and cover PYD-012

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#71**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Pydantic AI had no observability rule; OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern. The framing is Pydantic AI's own rather than a port: Pydantic AI emits OpenTelemetry spans for each run and tool call, so everything around this print is already correlated to a trace — the bare print is the one diagnostic landing outside it, unattached to the run that produced it.

## What this PR does

| case | expectation |
|---|---|
| `pprint({...})` | silent |

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "PYD-012 fires on print() in the tool body", ruleID: "PYD-012", kind: models.KindPydanticAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    print("looking up " + order_id)
+    return order_id
+`, wantFires: true},
+	{name: "PYD-012 silent when logging instead of printing", ruleID: "PYD-012", kind: models.KindPydanticAITool, src: `
+import logging
+logger = logging.getLogger(__name__)
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    logger.info("looking up %s", order_id)
+    return order_id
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks, @bradAGI. Pairing the fixture mirror with the tests in one PR, and saying plainly that neither half merges alone, is exactly the right shape. The fixture copy is byte identical to the file in trustabl/trustabl-rules#71, so `rules-sync` has nothing to object to once both land.

I walked the three cases against `PredHasPrintCall`. The fire case hits the bare `print` callee inside the discovered function node. The logger case resolves to an `attribute` callee (`logger.info`), and the module level `getLogger` sits outside the function node anyway, so it stays silent for the right reason. The `pprint` case resolves to the identifier `pprint`, which is not `print`. Nothing passes vacuously here, and I agree the third case is worth keeping as a standing guard on the bare callee contract rather than a comment.

The line removed from `policyRepoRuleCases` is the `gofmt` fix for that block, so that is a real cleanup rather than stray whitespace.

Only nit, and it is not blocking: the new cases land at the end of `policyRuleCases` rather than in the `PYD-*` banner block near `PYD-001` through `PYD-007`, and the file otherwise groups by SDK.

`PYD-012` still needs a rationale doc in `trustabl-rulebook` before it ships. That part is on us, and we are happy to pair on it.

PR #151, opened 2026-08-24
test(rules): mirror and cover LC-020

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#72**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

LangChain had no observability rule; OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern. The framing is LangChain's own rather than a port: LangChain already threads every tool start, end, and error through its callback system into LangSmith or whichever tracer is configured, so a bare print is the one diagnostic landing outside that — absent the moment anyone debugs from a trace rather than a terminal.

## What this PR does

| case | expectation |
|---|---|
| `pprint({...})` | silent |

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "LC-020 fires on print() in the tool body", ruleID: "LC-020", kind: models.KindLangChainTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    print("looking up " + order_id)
+    return order_id
+`, wantFires: true},
+	{name: "LC-020 silent when logging instead of printing", ruleID: "LC-020", kind: models.KindLangChainTool, src: `
+import logging
+logger = logging.getLogger(__name__)
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    logger.info("looking up %s", order_id)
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy of `langchain/observability.yaml` is byte identical to the one in trustabl/trustabl-rules#72, so `rules-sync` has nothing to object to, and sending the two halves on a branch of the same name is exactly the right way to do this.

The cases hold up on a read. `parsePy` takes the first `function_definition` in the snippet, so the module level `logger = logging.getLogger(...)` assignment and the `pprint` import both sit outside the parsed tool without changing which function gets picked, and `PredHasPrintCall` walks only that function node. The third case is a genuinely useful standing guard rather than padding, since the bare callee behavior is the whole reason this rule is not a `has_body_text` match.

One small thing before it lands: `COVERAGE.md` enumerates the LangChain Python tool rules by ID on the coverage row (`LC-001` through `LC-006`), so it now understates what ships. Add `LC-020` there in the same change and touch the `_Last reviewed_` line. The stray blank line drop in `policyRepoRuleCases` is unrelated but harmless, leave it.

`LC-020` will also want a rulebook rationale doc before it ships, and we can take that part. Add the coverage line and I am happy to take the pair.

PR #152, opened 2026-08-24
test(rules): mirror and cover CREW-012

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#73**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

CrewAI had no observability rule; OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern. The framing is CrewAI's own rather than a port: A crew running with verbose=True is already writing its own narration to stdout, so a tool print is interleaved into commentary from several agents at once with nothing marking which agent, task, or tool call emitted it.

## What this PR does

| case | expectation |
|---|---|
| `pprint({...})` | silent |

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "CREW-012 fires on print() in the tool body", ruleID: "CREW-012", kind: models.KindCrewAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    print("looking up " + order_id)
+    return order_id
+`, wantFires: true},
+	{name: "CREW-012 silent when logging instead of printing", ruleID: "CREW-012", kind: models.KindCrewAITool, src: `
+import logging
+logger = logging.getLogger(__name__)
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks for splitting this correctly, @bradAGI. The fixture copy of `crewai/observability.yaml` is byte identical to the one in trustabl/trustabl-rules#73, which is exactly what `rules-sync` wants, and the shared branch name means the cross-repo job resolves the matching pack rather than `main`.

The three cases hold up when I walk them against the predicate. `parsePy` takes the first `function_definition` in the snippet, so the module level `import logging` and `logger = ...` lines in the silent case do not confuse the harness, and `PredHasPrintCall` compares the callee text to exactly `print`, which makes the `pprint` case a real guard rather than a pass by accident. Adding that third case was the right call; it pins behavior that `schema.yaml` currently only describes in prose. The unrelated blank line dropped from `policyRepoRuleCases` is harmless, leave it in.

The one outstanding item is on our side, not yours: before CREW-012 ships it needs a rationale doc in `trustabl-rulebook` with the threat model and OWASP references, and we can handle that or pair with you on it. Since neither half stands alone, I will take this together with trustabl/trustabl-rules#73.

PR #153, opened 2026-08-24
test(rules): mirror and cover AG2-018

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#74**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

AutoGen had no observability rule; OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern. The framing is AutoGen's own rather than a port: The conversation itself is narrated to stdout, so a tool print lands mid-transcript and reads as though it were part of the exchange — when it is invisible to the agents reasoning over it.

## What this PR does

| case | expectation |
|---|---|
| `pprint({...})` | silent |

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "AG2-018 fires on print() in the tool body", ruleID: "AG2-018", kind: models.KindAutoGenTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    print("looking up " + order_id)
+    return order_id
+`, wantFires: true},
+	{name: "AG2-018 silent when logging instead of printing", ruleID: "AG2-018", kind: models.KindAutoGenTool, src: `
+import logging
+logger = logging.getLogger(__name__)
+def lookup_order(order_id: str) -> str:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. I reviewed this together with trustabl/trustabl-rules#74, and the fixture copy is byte identical to the production pack, so the pair keeps `rules-sync` green as long as they merge together (and neither should land alone, as your PR body already says). The test cases are the part I want to call out. The `pprint` case is a genuinely useful pin, since `PredHasPrintCall` deliberately matches only the bare `print` callee and the schema comment names substring matching as the exact trap. I walked all three cases against the predicate on main: the fire case is a bare print call inside the tool body, the logging case is an attribute callee so it stays silent, and `pprint` is a different bare callee, so all three exercise real behavior rather than passing vacuously. Using the prescribed remediation (a module logger) as the silent input, instead of just deleting the call, is a nice touch too. The only stray bit is the unrelated blank-line removal at the end of `policyRepoRuleCases`, which is harmless. As noted on the rules half, AG2-018 will need a rationale doc in trustabl-rulebook before it ships; the maintainers can take that or pair on it. Approving, and I will coordinate landing the two halves together.

PR #154, opened 2026-08-24
test(rules): mirror and cover MCP-201

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#78**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

**MCP was the only pack with no repo-scope rule at all** — the `mcp` `applies_to` token was valid in `appliesToByScope` but no shipped rule used it. Every other pack ships the agent-guidance check (CSDK-203, OAI-202, ADK-201, LC-201, CREW-201, AG2-201, PYD-201, VAI-012).

The framing is MCP's own: everything the repo exports crosses a trust boundary, so a tool added without the local conventions doesn't fail *there* — it degrades tool selection in every downstream client, and the server's authors aren't the ones who observe it.

## What this PR does

1. Mirrors `mcp/repo_hygiene.yaml` into `testdata/rules-fixture/`.
2. Adds cases to `policyRepoRuleCases`, as `TestPolicyRules_AllRulesCovered` requires.

| case | expectation |
|---|---|
| MCP code detected, no AGENTS.md or CLAUDE.md | fires |
| same, with an AGENTS.md component | silent |
| **LangChain detected instead of MCP** | silent |

Three cases, matching the CSDK-203 / ADK-201 / OAI-202 group this rule joins. The third pins that the `repo_has_sdk_in_code` gate actually gates — a repo-scope rule with an over-broad match fires once per scan on *every* repo, so that case is the one keeping it honest.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.082s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -2247,6 +2247,25 @@ var policyRepoRuleCases = []policyRepoCase{
 		models.RepoInventory{SDKsDetected: []models.SDK{models.SDKOpenAIAgents}},
 		false},
 
+	// ─── MCP-201: MCP server repo with no agent-guidance doc ────────────────
+	// Same shape as CSDK-203 / ADK-201 / OAI-202: repo_has_sdk_in_code reads
+	// inv.SDKsDetected, repo_component_present reads
+	// profile.Manifest.Components. Either AGENTS.md or CLAUDE.md silences it.
+	{"MCP-201 fires when MCP server code has no agent-guidance doc", "MCP-201",
+		models.RepoProfile{Languages: []models.Language{models.LanguagePython}},
+		models.RepoInventory{SDKsDetected: []models.SDK{models.SDKMCP}},
+		true},
+	{"MCP-201 silent when AGENTS.md is present", "MCP-201",
+		models.RepoProfile{
+			Languages: []models.Language{models.LanguagePython},
+			Manifest:  models.ScanManifest{Components: []models.AgentComponent{{Kind: models.ComponentAgentsMd}}},
+		},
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy of `mcp/repo_hygiene.yaml` is byte-identical to the one in trustabl/trustabl-rules#78 (same blob hash), so `check-rules-sync.sh` has nothing to complain about, and pairing the two halves on a same-named branch is exactly how I want these submitted.

The test cases hold up when I walk them. The fire case leaves `Manifest.Components` empty, so `repo_component_present` is false and the `not` holds, while `repo_has_sdk_in_code: mcp` matches `SDKsDetected`, so `MCP-201` fires. The `ComponentAgentsMd` case flips the second leg and silences it.

One correction on the description, not on the code. The third case does not pin the `repo_has_sdk_in_code` gate. `TestPolicyRepoRules` calls `Applies` first, and at repo scope `Applies` already filters on `applies_to` against `SDKsDetected`, so a LangChain-only inventory returns before the match expression is evaluated at all. The case is still worth keeping (it pins the `applies_to` gate, which is the other thing that would make a repo rule fire on every scan), just not for the stated reason. No change needed.

The rule still needs a rationale doc in `trustabl-rulebook` before it ships, which is on the maintainer side. Merging alongside trustabl/trustabl-rules#78.

sairenchristianbuerano:

Nothing needed from you.

sairenchristianbuerano (approved):

PR #155, opened 2026-08-24
test(rules): mirror and cover ADK-112

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#79**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

`adk_sequential_agent`, `adk_parallel_agent`, and `adk_langgraph_agent` are valid `applies_to` tokens the engine discovers but **no shipped rule used any of them** — every ADK agent-scope rule targets `adk_llm_agent`, except ADK-108 for `LoopAgent`. I found this auditing which engine capabilities have zero rule coverage.

ADK-112 extends ADK-101's routing check to the workflow classes: a workflow agent is routed to exactly like an LlmAgent (the parent's model sees only the child's `description`), but it's easier to miss because a workflow agent has no `instruction=` or `model=` of its own, so nothing else about its construction reads as model-facing text.

## What this PR does

1. Mirrors `google_adk/agent_safety.yaml` into `testdata/rules-fixture/`.
2. Adds cases to `policyAgentRuleCases`, as `TestPolicyRules_AllRulesCovered` requires.

| case | expectation |
|---|---|
| `SequentialAgent` with no `description` | fires |
| `ParallelAgent` with no `description` | fires |
| `LoopAgent` with no `description` | fires |
| `SequentialAgent` with a `description` | silent |

**One fire case per class rather than a single representative.** The three reach the rule through different `applies_to` tokens sharing one match, so the per-class cases are what would catch a token being dropped from the list — a single case would leave two thirds of the rule's reach untested.

`adk_langgraph_agent` is deliberately excluded from the rule: `LanggraphAgent` takes a compiled graph rather than an ADK `sub_agents` tree, so the delegation-routing argument doesn't obviously transfer. Noted in the rules PR.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.019s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -2246,7 +2246,6 @@ var policyRepoRuleCases = []policyRepoCase{
 		},
 		models.RepoInventory{SDKsDetected: []models.SDK{models.SDKOpenAIAgents}},
 		false},
-
 }
 
 // optionsWithPermissionMode builds a ClaudeAgentOptionsDef whose captured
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy matches the production YAML in trustabl/trustabl-rules#79 line for line, which is the part of a cross-repo pair that usually goes wrong, and your read of the `rules-sync` job is right: it resolves `head_ref` against the rules repo before falling back to `main`, so the pair validates before either side merges. I have reviewed both halves and they should land together.

The tests are the part I liked most. One fire case per `applies_to` token is the correct instinct rather than overkill, because `agentKindMatches` keys off `SDK` plus `Class`, so a token dropped from the list would leave a whole class silently unmatched and a single representative case would sail past it. The silent case carries a real `description` string, so the rule genuinely evaluates instead of falling out at `Applies` and passing vacuously, which is the usual way a silent case goes hollow.

Minor, and not a blocker: the blank-line removal in `policyRepoRuleCases` at the top of the diff is unrelated to `ADK-112` and only widens the change, so I would drop it if you touch the branch again.

Same note as on the rules PR, the rule needs a rationale doc in `trustabl-rulebook` before it ships, and that part is ours. I am happy to take both halves together.

sairenchristianbuerano:

Nothing needed from you.

sairenchristianbuerano (approved):

PR #156, opened 2026-08-24
test(rules): mirror and cover MCP-023

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#82**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

MCP had no observability rule. OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern — both at `low`, framed as a lost-diagnostic problem.

**It isn't a lost-diagnostic problem here, which is why MCP-023 ships at `medium`.** On a stdio server — the FastMCP default, and how an editor or desktop client launches one — stdout *is* the protocol channel. A loose print interleaves with the newline-delimited JSON-RPC frames, so the client hits a parse error on a line that isn't JSON and drops the response or tears down the session. The symptom doesn't resemble the cause: a tool call that returns nothing, traced back to a print that reads like harmless debugging.

OAI-010's explanation already anticipates exactly this case as a footnote; MCP-023 makes it the primary reading. Confidence 0.7 because HTTP/SSE transports escape the corruption (though still lose the diagnostic).

## What this PR does

1. Mirrors `mcp/observability.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| `logger.info(...)` (stderr handler) | silent |
| `pprint({...})` | silent |

The `pprint` case pins `has_print_call`'s bare-callee behavior — the rule must not regress into substring matching that sweeps in `pprint` and every other callee whose name merely contains "print". The schema calls that out as the trap `has_body_text` falls into, so it's worth a standing test rather than a comment.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
ok  	github.com/trustabl/trustabl/internal/rules	3.525s
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,27 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "MCP-023 fires on print() in the tool body", ruleID: "MCP-023", kind: models.KindMCPTool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    print("looking up " + order_id)
+    return order_id
+`, wantFires: true},
+	{name: "MCP-023 silent when logging to stderr instead", ruleID: "MCP-023", kind: models.KindMCPTool, src: `
+import logging
+logger = logging.getLogger(__name__)
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    logger.info("looking up %s", order_id)
+    return order_id
+`, wantFires: false},
+	{name: "MCP-023 silent on pprint (bare-callee guard)", ruleID: "MCP-023", kind: models.KindMCPTool, src: `
+from pprint import pprint
+def lookup_order(order_id: str) -> str:
+    """Look up an order."""
+    pprint({"order_id": order_id})
+    return order_id
+`, wantFires: false},
 }
 
 // policyRepoRuleCases covers repo-scoped rules.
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is a clean half of a clean pair. The fixture copy of `mcp/observability.yaml` is byte for byte identical to the file in trustabl/trustabl-rules#82, so `check-rules-sync.sh` has nothing to complain about, and the three cases follow the existing MCP conventions in `policyRuleCases` exactly, `toolConfig` shape included.

Two small things before I take it. The cases sit after the `CSDK-018` block at the end of the slice rather than in the MCP pack section further up, so please move them there with a section header like the neighbours have. And `COVERAGE.md` still lists the MCP Python rules as `MCP-001..010`, so add `MCP-023` to that cell (the narrative section lower in that file is already behind, leave that part alone).

On the `pprint` case, `has_print_call`'s bare callee behaviour is already pinned for `OAI-010` higher up in the same slice, so this is belt and braces, but I am fine carrying a second one on the MCP side. The stray whitespace change in `policyRepoRuleCases` is harmless.

This needs to land together with trustabl/trustabl-rules#82, never on its own. `MCP-023` also needs a rationale doc in `trustabl-rulebook` before it reaches users; that one is ours, and I am glad to pair on it. Happy to take this once the case placement and the `COVERAGE.md` line are in.

PR #158, opened 2026-08-24
test(rules): mirror and cover CSDK-020

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#84**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

CSDK-003 covers the Python side of network timeouts; the TypeScript half was missing, even though this pack ships TS rules throughout (CSDK-010..014, CSDK-120..131). OpenAI (OAI-016, OAI-024) and Vercel AI (VAI-011) already use `has_http_call_without_timeout` for exactly this.

The consequence framing is what makes it more than a port of VAI-011: **an unresponsive host stalls the conversation rather than failing it.** The model gets no result and no error, so the turn can't advance — and a `max_turns` cap doesn't help, because the run is stuck *inside* one turn rather than taking too many.

## What this PR does

1. Mirrors `claude_sdk/network.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

Three cases, following OAI-016's own table. The third is the one worth having: it pins that the predicate checks for a **deadline**, not merely for an options argument — the plausible regression is treating any second argument as "configured".

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,37 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── CSDK-020: TS Claude SDK tool HTTP call without a timeout ───────────
+	// Structural has_http_call_without_timeout, as OAI-016 uses. CSDK-003 is
+	// the Python half; this pack shipped TS rules but no TS timeout rule.
+	{
+		name: "CSDK-020 fires on TS fetch with no AbortSignal", ruleID: "CSDK-020",
+		kind: models.KindClaudeSDKTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { tool } from \"@anthropic-ai/claude-agent-sdk\";\n" +
+			"export const t = tool(\"fetch_report\", \"Fetch a report.\", {}, async () => {\n" +
+			"  const r = await fetch(\"https://reports.internal/x\");\n" +
+			"  return { content: [{ type: \"text\", text: await r.text() }] };\n" +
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is the shape I want cross-repo rule work to take. I diffed the fixture hunk against trustabl/trustabl-rules#84 and the two are byte identical, so `check-rules-sync.sh` should be satisfied once both land together, and I agree neither half merges alone.

The test cases fit the harness correctly. `models.KindClaudeSDKTool` with `models.LanguageTypeScript` routes through `DiscoverTSTools`, your snippets carry the `@anthropic-ai/claude-agent-sdk` import that gates discovery and four positional args so extraction succeeds, and each case lines up with `httpCallHasTimeout`: no options object at all, an options object carrying `signal`, and an options object carrying only `method`. The third one is the case I would have asked for, since the plausible regression is treating any second argument as configured. Fire and silent are both present, so the coverage guard is satisfied. The blank line dropped from `policyRepoRuleCases` is unrelated but harmless.

I left two small text nits on trustabl/trustabl-rules#84, one of which changes the explanation prose. When you apply it there, mirror the identical edit into `testdata/rules-fixture/claude_sdk/network.yaml` here, or the sync check will fail on the drift. The rule also needs a rationale doc in `trustabl-rulebook` before it ships; that part is on us and I am glad to pair on it. Happy to take this once the fixture matches the updated rules half.

PR #159, opened 2026-08-24
test(rules): mirror and cover MCP-024

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#85**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

MCP-004 covers the Python side of network timeouts; the TypeScript half was missing, even though the pack ships TS rules (MCP-011, MCP-013).

**What makes the MCP case its own is that the stall crosses the server's trust boundary.** The connecting client waits on a JSON-RPC response that never arrives, and MCP gives it no way to cancel an in-flight tool call — so it's left to its own timeout, an abandoned request, or a hung session, with nothing saying why. The server author never sees the symptom; the client's users do. On a stdio server a handler parked on a dead socket is also holding the single process that serves the connection.

## What this PR does

1. Mirrors `mcp/network.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

Three cases, following OAI-016's own table. The third pins that the predicate checks for a **deadline**, not merely for an options argument.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,38 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── MCP-024: TS MCP tool HTTP call without a timeout ───────────────────
+	{
+		name: "MCP-024 fires on TS fetch with no AbortSignal", ruleID: "MCP-024",
+		kind: models.KindMCPTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { McpServer } from \"@modelcontextprotocol/sdk/server/mcp.js\";\n" +
+			"const s = new McpServer({ name: \"orders\", version: \"1.0.0\" });\n" +
+			"s.tool(\"fetch_report\", \"Fetch a report.\", {}, async () => {\n" +
+			"  const r = await fetch(\"https://reports.internal/x\");\n" +
+			"  return { content: [{ type: \"text\", text: await r.text() }] };\n" +
+			"});\n",
+	},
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the shape a coordinated change should arrive in: the `mcp/network.yaml` block added here is byte identical to the one in trustabl/trustabl-rules#85, and both halves sit on the same branch name, so the `rules-sync` job resolves the matching production pack rather than `main`.

The three cases hold up when I walk them against the predicate. `has_http_call_without_timeout` reads the discovery fact `http_no_timeout`, and since `parseTSTool` runs the real `DiscoverTSMCPProper`, the facts come from `tsHandlerFacts` on the actual handler node instead of being hand stuffed into `Facts`, so none of these pass vacuously. The bare `fetch` case sets the fact, the `signal: AbortSignal.timeout(...)` case clears it through `httpCallHasTimeout`, and the `{ method: "POST" }` case is the one that earns its place, because it pins that the check is for a deadline and not merely for the presence of an options argument. Leaving `schema_version` alone is correct; the predicate has been on `main` since schema 14.

One nit: the diff also drops a blank line in `policyRepoRuleCases`, unrelated to this change. Harmless, no need to revert it.

MCP-024 still needs a rationale doc in `trustabl-rulebook` before it ships, but that is our side of the work and we are happy to pair on it. I will take this together with the rules half.

PR #160, opened 2026-08-24
test(rules): mirror and cover ADK-114

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#86**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

ADK-003 covers the Python side of network timeouts; the TypeScript half was missing, even though the pack ships TS rules (ADK-013, ADK-015, ADK-016, ADK-109).

**The composition ADK encourages makes the stall worse rather than better:** inside a `SequentialAgent` a stalled tool blocks every step after it, and inside a `ParallelAgent` the fan-out waits on its slowest branch. And the limit people reach for doesn't apply — ADK-108's `max_iterations` bounds how many times a `LoopAgent` goes round, not how long one tool call may run.

## What this PR does

1. Mirrors `google_adk/network.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

Three cases, following OAI-016's own table; the third pins that the predicate checks for a **deadline**, not merely for an options argument.

**One note on fixture shape.** The snippets use adk-js's options-object form (`new FunctionTool({ ..., execute })`). My first draft used the Python-style `new FunctionTool(fn, {...})` wrapper and *nothing* fired — including the pre-existing TS rules. Worth flagging because the Python `FunctionTool(fn)` shape reads as the obvious analogue and isn't what the TS SDK does.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,37 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── ADK-114: TS ADK FunctionTool HTTP call without a timeout ───────────
+	// adk-js uses the options-object form (new FunctionTool({ ..., execute })),
+	// not the Python FunctionTool(fn) wrapper shape.
+	{
+		name: "ADK-114 fires on TS fetch with no AbortSignal", ruleID: "ADK-114",
+		kind: models.KindADKFunctionTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { FunctionTool } from \"@google/adk\";\n" +
+			"const t = new FunctionTool({ name: \"fetch_report\", description: \"Fetch a report.\", parameters: {}, execute: async () => {\n" +
+			"  const r = await fetch(\"https://reports.internal/x\");\n" +
+			"  return await r.text();\n" +
+			"} });\n",
+	},
+	{
+		name: "ADK-114 silent when AbortSignal present", ruleID: "ADK-114",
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture mirror is byte-identical to the YAML in trustabl/trustabl-rules#86, so `check-rules-sync.sh` has nothing to complain about once both land together, and the three cases sit in `policyRuleCases` exactly where `TestPolicyRules_AllRulesCovered` looks for them.

I walked the cases against the predicate rather than trusting the shape. `parseTSTool` runs the real ADK TS discovery, so the facts are computed from the AST rather than stubbed. The bare `fetch` and the `{ method: "POST" }` case both leave `http_no_timeout` set, because `httpCallHasTimeout` only accepts an options object carrying `signal`, `timeout`, or `abortSignal`, and the `AbortSignal.timeout(15000)` case clears it via the `signal` key. Adding the third case is the right instinct: it pins that the predicate checks for a deadline and not merely for the presence of an options argument. The import gate and the `new FunctionTool({ ..., execute })` shape match the existing ADK-013, ADK-015, and ADK-016 cases, and your note about the Python wrapper shape is worth keeping in the file comment for the next person.

The unrelated blank-line removal in `policyRepoRuleCases` is harmless. Nothing blocking here. I have reviewed the rules half alongside this one, and the two should merge together.

PR #161, opened 2026-08-24
fix(cli): stop claiming JavaScript and Go are recon-only

Paramjeet-singh-neu:
## Summary
- The root command `Long` help still said JavaScript and Go are recognized during recon but **not AST-parsed**.
- That has been wrong since the TS-family JS pipeline and Go/C#/PHP/Rust MCP discovery landed. `trustabl --help` was misinforming users.

One file, one stale sentence. No rule IDs.

## Test plan
- [x] `trustabl --help` mentions JavaScript in depth and MCP discovery for Go/C#/PHP/Rust
- [x] Help no longer says JS/Go are recon-only

### DIFF
--- a/cmd/trustabl/main.go
+++ b/cmd/trustabl/main.go
@@ -75,9 +75,9 @@ func main() {
 It scans repositories that use agent SDKs — Claude Agent SDK, OpenAI Agents SDK,
 Google ADK, LangChain, CrewAI, Pydantic AI, Vercel AI, and AutoGen — and Model
 Context Protocol (MCP) servers, then reports reliability and safety weaknesses in
-the tools, agents, and subagents it discovers. Python and TypeScript codebases
-are analyzed in depth; JavaScript and Go are recognized during recon but not yet
-AST-parsed.
+the tools, agents, and subagents it discovers. Python, TypeScript, and JavaScript
+are analyzed in depth (JS through the TypeScript-family pipeline). Go, C#, PHP,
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu, good catch and the replacement text is accurate. The old sentence has been wrong since the JS and MCP-language work landed. In `scanner.Run` the TS block appends `Manifest.JavaScriptFiles` into the same parse batch as the TypeScript files, runs every TS-family discovery pass over them, and `retagJavaScriptDefs` then stamps the results `javascript` after edge resolution, so JavaScript really is analyzed in depth rather than recon-only. Go, C#, PHP, and Rust each get a dedicated tree-sitter parse pass feeding `DiscoverGoMCPTools`, `DiscoverCSharpMCPTools`, `DiscoverPHPMCPTools`, and `DiscoverRustMCPTools`, which is exactly the MCP-server scope your new sentence claims and nothing wider. The wording also matches what `README.md` and `COVERAGE.md` already say, so the docs stay consistent.

I also checked the rest of the tree and `cmd/trustabl/main.go` was the only place still carrying the stale claim. The similar lines in `README.md` and `ARCHITECTURE.md` are about the non-MCP Go, .NET, PHP, and Rust SDKs and remain correct in context, so nothing else needs to move with this. Happy to take it as is.

sairenchristianbuerano (approved):

PR #162, opened 2026-08-24
fix(cli): llm help still described enrichment as a future path

Paramjeet-singh-neu:
## Summary
- `trustabl llm --help` still said keys were only stored for a **future path** and that enrichment was not wired.
- `trustabl enrich` already uses that BYOK config. Scan still never calls an LLM — the help now says that without pretending enrich does not exist.

Independent of the root-help PR. No rule IDs.

## Test plan
- [x] `go test ./cmd/trustabl/ -run TestLLMCommand_HelpDescribesEnrichment`
- [ ] `trustabl llm --help` mentions `trustabl enrich` and does not say "future path"

### DIFF
--- a/cmd/trustabl/llm.go
+++ b/cmd/trustabl/llm.go
@@ -18,12 +18,12 @@ func newLLMCommand() *cobra.Command {
 		Use:   "llm",
 		Short: "Manage LLM provider configuration",
 		Long: `Manage the optional LLM provider configuration (provider, model, and API key)
-used for bring-your-own-key enrichment. Configuration is stored in your user
-config directory with 0600 permissions.
+used by "trustabl enrich" for bring-your-own-key enrichment. Configuration is
+stored in your user config directory with 0600 permissions.
 
-NOTE: LLM enrichment is not wired into the scan yet. These commands only store
-the provider, model, and key for that future path — no scan makes an LLM call
-today, with or without a key configured.`,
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu, this is a real fix and you spotted it correctly. The old note predates `trustabl enrich`, and `cmd/trustabl/enrich.go` on main does load the BYOK config and refuses to run without a key, so the shipped help genuinely misleads today. I also like that you kept the honest half (scan itself never makes an LLM call) rather than just deleting the caveat, and that the test pins both the removed phrases and the two command names so it cannot quietly regress.

One thing I want fixed before I take it: the precedence is stated backwards. `llm.Load` in `internal/llm/llm.go` checks `ANTHROPIC_API_KEY`, then `OPENAI_BASE_URL`, then `OPENAI_API_KEY`, then `GOOGLE_API_KEY`, and returns on the first hit with a fresh default config, so an exported key overrides the stored provider entirely instead of acting as a fallback for it. Read as written ("uses this stored provider, or ... from the environment"), someone with `ANTHROPIC_API_KEY` exported who runs `trustabl llm provider set openai` would expect OpenAI and silently get Anthropic. Please word it so the environment is checked first and wins. Worth naming `OPENAI_BASE_URL` in the same sentence, since it takes the custom-endpoint path even with no key set.

Happy to merge once the precedence wording is corrected.

sairenchristianbuerano (approved):

PR #163, opened 2026-08-24
docs(telemetry): align TELEMETRY.md with events the CLI actually sends

Paramjeet-singh-neu:
## Summary
TELEMETRY.md claims to be the authoritative event schema and says stale docs are a defect. It had drifted:

- `schema_version` example was `4` (pack / engine is **14**)
- `cli_version` / crash `version` example was `0.9.1` (latest release is **0.1.7**)
- `sdks_detected` example used detector category names (`openai_sdk`, `claude_sdk`); `scan.completed` emits `models.SDK` values (`openai_agents`, `claude_agent_sdk`, …)
- `command.run` omitted `forge`
- `scan.failed` omitted `exit_code`, which the CLI always sends as `2`

README points here as the contract. Independent of the CLI help PRs.

## Test plan
- [ ] Skim `cmd/trustabl/scan.go` / `forge.go` Track() calls against the updated tables

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -60,7 +60,7 @@ These properties are merged into every event automatically:
 | Property | Type | Example | Purpose |
 |---|---|---|---|
 | `anonymous_id` | string | `"a3f2c…"` | Anonymous installation identifier (UUID v4) |
-| `cli_version` | string | `"0.9.1"` | Correlate observations to releases |
+| `cli_version` | string | `"0.1.7"` | Correlate observations to releases |
 
 ---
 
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu, this is a useful cleanup and I checked each claim against the code on `main`. The `models.SDK` set really is `claude_agent_sdk`, `openai_agents`, `mcp`, `google_adk`, `langchain`, `crewai`, `pydantic_ai`, `vercel_ai`, `autogen`, and `scan.completed` stringifies `result.SDKs` directly, so the old `openai_sdk` / `claude_sdk` example was wrong in exactly the way you describe. `cmd/trustabl/scan.go` does send `exit_code: 2` on `scan.failed`, and documenting it also makes the minimal level row honest, since minimal mode forwards `exit_code` when it is present. `cmd/trustabl/forge.go` sends `command: "forge"` plus `explicit_count`, and `14` and `0.1.7` match `SupportedSchemaVersion` and the latest tag.

Two small things, neither blocking. Pinning `0.1.7` and `14` as examples means they drift again on the next release or schema bump. The parenthetical pointing at `manifest.yaml` and `SupportedSchemaVersion` helps a lot, and the same treatment on `cli_version` would be nice. Separately, `command.run` still says it fires for every non-scan subcommand, but `trustabl telemetry ...` and `trustabl forge check` do not emit it. That one is pre-existing, so take it here or leave it, your call.

Happy to take this as it stands.

sairenchristianbuerano (approved):

PR #164, opened 2026-08-24
test(rules): mirror and cover OAI-025, OAI-026

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#87**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

Ports the CSDK-017/018 description-quality pair to the OpenAI Agents SDK. OAI-001 only checks that a docstring *exists*, so `"""TODO: describe this tool."""` and `"""Gets data."""` pass today. Two OpenAI-specific points, both aimed at mitigations people assume already cover this:

- **Strict schemas don't compensate.** OAI-003's `strict_mode` constrains the shape of the arguments, never whether calling this tool was the right move — a placeholder description yields a well-formed call to the wrong tool.
- **With handoffs, the blast radius is the whole conversation.** An agent picks between its own tools and its peers' from these strings, so it can route to the wrong *agent*, not just the wrong function.

## What this PR does

1. Mirrors `openai_sdk/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|
| OAI-025 — `"""TODO: describe this tool."""` | fires |
| OAI-025 — real description | silent |
| OAI-026 — `"""Gets data."""` | fires |
| OAI-026 — full sentence | silent |
| OAI-026 — **no docstring at all** | silent |

**Five cases rather than four.** OAI-026 pairs `description_length_lt: 40` with `has_docstring: true` so an absent docstring stays OAI-001's finding rather than double-reporting — an empty description is length 0, which is also under the threshold. The fifth case pins that guard.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,30 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "OAI-025 fires on placeholder description", ruleID: "OAI-025", kind: models.KindOpenAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """TODO: describe this tool."""
+    return order_id
+`, wantFires: true},
+	{name: "OAI-025 silent on a real description", ruleID: "OAI-025", kind: models.KindOpenAITool, src: `
+def lookup_order(order_id: str) -> str:
+    """Look up a single order by its identifier and return its current status."""
+    return order_id
+`, wantFires: false},
+	{name: "OAI-026 fires on a too-short description", ruleID: "OAI-026", kind: models.KindOpenAITool, src: `
+def list_orders(customer_id: str) -> str:
+    """Gets data."""
+    return customer_id
+`, wantFires: true},
+	{name: "OAI-026 silent on a full description", ruleID: "OAI-026", kind: models.KindOpenAITool, src: `
+def list_orders(customer_id: str) -> str:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean engine half. I compared the fixture hunk against the YAML in trustabl/trustabl-rules#87 and the two are identical, and the base file was already byte-for-byte in sync between the fixture and production, so `rules-sync` stays green as long as the two land together. Please do not merge this one on its own.

Both predicates the rules use, `has_description_text` and `description_length_lt`, already exist on `main` under schema version 14, so leaving `SupportedSchemaVersion` and both manifests alone was the right call.

The test cases hold up on a read. `parsePy` fills `ToolDef.Description` from the docstring, so the fire and silent cases exercise the real predicate path rather than passing vacuously, and the fifth case pinning the `has_docstring: true` guard on OAI-026 is exactly the case I would have asked for.

One small thing: the diff also drops a blank line at the end of `policyRepoRuleCases`, which is unrelated to this change. Not worth another round on its own, but worth removing if you push again for any other reason.

New rules also need a rationale doc in `trustabl-rulebook` before they ship. That part is on us rather than on you, and we are happy to pair on it.

PR #169, opened 2026-08-24
feat(analysis): make has_raise / has_try_except work on TypeScript tools

eulestadt:
`has_raise` matches the tree-sitter node type `raise_statement`, which doesn't exist in the TypeScript grammar — the equivalent there is `throw_statement`. So the predicate returns false for every TS tool, and `has_try_except` only works by accident, because `try_statement` happens to be spelled the same in both grammars.

The reason I think this is worth fixing before anyone writes the rule that needs it: a TypeScript error-contract rule would load, pass `rules validate`, show up in `POLICY_INDEX.md`, and never fire. That reads as coverage rather than as a gap, which is the failure mode the "unaudited SDK" info finding exists to prevent elsewhere in the pipeline. Four packs ship a Python error-contract rule today (CSDK-005, OAI-008, ADK-005, MCP-006) and none of them can have a TS counterpart until this lands.

This was spotted from the rules side in #55, with the suggestion that `PredHasShellCall` already branches on `models.IsTSOrJS` and reads a discovery-computed fact, so `PredHasRaise` could do the same. That's what this does. `tsHandlerFacts` records `throws` and `try_catch` alongside the `shells_out` / `code_exec` / `http_no_timeout` facts it already computes, and both predicates read them when the language is TS or JS. The Python paths are untouched. Since `tsHandlerFacts` is shared by all six TS discovery paths — Claude SDK, LangChain, OpenAI Agents, MCP, Vercel AI, ADK — this opens up TS error-contract rules in every pack at once instead of one at a time.

One semantic decision worth surfacing: a re-throw from inside a `catch` sets both facts, so the shipped `has_raise AND NOT has_try_except` shape stays silent on it. That matches what Python already does for a `raise` inside an `except`. I pinned it with a test rather than leaving it implicit, because scoping `throws` to exclude catch blocks would have made the two languages disagree on identical logic.

For tests there are five fact-level cases, plus predicate-level cases parameterized over all five TS tool kinds, so a regression in the shared fact walk shows up here rather than in whichever pack ships the first TS error rule.

The unit tests only prove the predicate, though, so I also checked it end to end. I wrote a throwaway `language: typescript` rule using `has_raise` + `not has_try_except`, then scanned a LangChain.js project with one throwing tool and one that catches and returns a structured error. Built from `main`, the rule resolves and evaluates and produces nothing:

```
  lookupOrder
    [LC-013] HIGH TypeScript LangChain tool fetches a caller-controlled URL (SSRF)
  (repo-wide)
    [LC-201]  LOW LangChain project ships no agent-guidance doc
```

With this branch:

```
  tool:lookupOrder    78%  (2 findings)
  tool:fetchInvoice   81%  (1 findings)

  lookupOrder
    [LC-013] HIGH TypeScript LangChain tool fetches a caller-controlled URL (SSRF)
    [LC-999]  LOW LangChain TS tool throws without a structured error contract  (src/tools.ts:4-12)
  fetchInvoice
    [LC-013] HIGH TypeScript LangChain tool fetches a caller-controlled URL (SSRF)
```

`fetchInvoice` keeps its one pre-existing finding and gains nothing, which is the control I cared about. The scratch `LC-999` rule lives only in that throwaway pack and isn't part of this PR.

`go test ./...`, `go vet` and `gofmt` are all clean, and `check-rules-sync.sh` passes because this touches no rules or fixture — which also means it merges on its own, unlike the paired rule PRs.

No version bump. There's no new predicate name and no schema field, so this widens an existing predicate rather than redefining one: no shipped rule pairs `has_raise` with `language: typescript`, so nothing changes behavior today, and an older binary reading a future TS error rule skips it on the `language` gate the way it does now. I updated the `DiscoverTSTools` fact list in `ARCHITECTURE.md` in the same commit — it was also missing `http_no_timeout`, `writes_fs` and `code_exec`.

I deliberately left the actual TS error-contract rules out of this. The primitive is worth landing on its own, and the per-SDK `explanation` and `fix` text should probably come from whoever's closest to each pack, since it has to name a framework-specific consequence — the LangChain executor stringifying an exception into a `ToolMessage` is a different story from an MCP client seeing a transport error.


jhumel-code:
Thanks @eulestadt, good catch and a clean fix. I confirmed `raise_statement` is absent from the TS grammar, and that all four shipped error-contract rules (`CSDK-005`, `OAI-008`, `ADK-005`, `MCP-006`) are explicitly `language: python`, so nothing on main changes behavior. Reading the fact from `tsHandlerFacts` rather than re-walking the tree inside the predicate is the right shape, and running the predicate tests through real discovery instead of hand-built `ToolDef`s is why I trust them.

Two asks. The comment on `tsErrorContractCases` says one snippet per discovery path and all six, but there are five: `KindADKFunctionTool` is missing, and `ts_adk_tools.go` reads its handler off the `execute` key, so it is worth pinning. `ARCHITECTURE.md` also has two stale enumerations this change invalidates: the `ts_handler_facts.go` line in the package tree still lists only `shells_out, writes_fs, http_call, dynamic_url`, and the `predicates.go` line still names only `PredHasShellCall`, `PredHasWriteCall`, `PredHasCodeExecCall` and `PredHasDynamicURLCall` as TS-aware.

One correction on the no-bump reasoning. An older binary reading a future `language: typescript` rule that uses `has_raise` does not skip it: the language gate passes, the rule evaluates, and it returns false, so it silently never fires. I agree no bump is needed here, but let us decide that deliberately when the first TS error rule lands rather than inherit it.

Also worth flagging for whoever writes that rule: `try_statement` covers `try/finally` with no `catch` in both grammars, so `try_catch` will silence a handler that does not actually handle anything. Parity with Python, but it belongs in the rulebook rationale doc, which the maintainers can pair on when the rule itself lands.

Happy to take this once the ADK case and those two `ARCHITECTURE.md` lines are in.

sairenchristianbuerano (approved):

PR #172, opened 2026-08-24
test(rules): mirror and cover MCP-029

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#94**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

MCP-005 covers the Python path-safety case; the TypeScript half was missing. MCP-029 mirrors CSDK-012 including its coarse-signal caveat, stated in the explanation: it flags *any* filesystem write, not only unnormalized paths, because TS path-normalization analysis isn't wired yet. Confidence 0.5 to match.

**Deployment is what sharpens this for MCP.** A stdio server is launched as a subprocess by whatever client the user is running, so it inherits **that user's own filesystem permissions** — a write escaping its intended directory reaches their home directory, dotfiles, and SSH keys. And the server can't see the injection: it receives a well-formed `tools/call` for a path it has no way to distinguish from a legitimate one, so containment has to be structural.

## What this PR does

1. Mirrors `mcp/path_safety.yaml` into `testdata/rules-fixture/`.

The silent case applies the remediation the `fix` text prescribes for the common shape — **derive the filename server-side rather than accepting a path over the protocol** — instead of merely deleting the write.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,32 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── MCP-029: TS MCP tool writes to the filesystem ──────────────────────
+	// Coarse has_write_call signal, mirroring CSDK-012 until TS path
+	// normalization analysis exists.
+	{
+		name: "MCP-029 fires on filesystem write", ruleID: "MCP-029",
+		kind: models.KindMCPTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { McpServer } from \"@modelcontextprotocol/sdk/server/mcp.js\";\n" +
+			"import { writeFileSync } from \"node:fs\";\n" +
+			"const s = new McpServer({ name: \"notes\", version: \"1.0.0\" });\n" +
+			"s.tool(\"save_note\", \"Save a note to disk for later retrieval.\", {}, async ({ p, body }) => {\n" +
+			"  writeFileSync(p, body);\n" +
+			"  return { content: [] };\n" +
+			"});\n",
+	},
+	{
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. Pairing the fixture mirror with the rules half on a same-named branch is the right way to do this, and the mirrored block in `testdata/rules-fixture/mcp/path_safety.yaml` is byte identical to trustabl/trustabl-rules#94, so `rules-sync` has nothing to complain about once both land.

The test cases are real rather than vacuous. Both snippets import from `@modelcontextprotocol/sdk/server/mcp.js` and bind `s` to `new McpServer(...)`, so `DiscoverTSMCPProper` genuinely produces a `KindMCPTool` in each case. The fire case's `writeFileSync` is one of the callees `tsHandlerFacts` maps to `writes_fs`, and the silent case still discovers a tool but sets no fact, so it exercises the predicate rather than the import gate. Writing the silent case as the remediation the `fix` text prescribes, instead of just deleting the write, is a nice touch.

Two things. The diff also deletes a blank line at the end of `policyRepoRuleCases`, which is unrelated to this change and will conflict with other open PRs touching that file, so please drop it. And `MCP-029` is contested: trustabl/trustabl#124 and trustabl/trustabl#135 both claim it for different rules. Main stops at `MCP-022`, so renumbering is cheap for whoever lands second.

Happy to take this once those two are handled.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `MCP-029`.

Closing trustabl/agent-reliability-rules#80 (PHP typed parameters) and trustabl/agent-reliability-rules#100 (PHP idempotency), which claimed the same id for different rules.

PR #174, opened 2026-08-24
test(rules): mirror and cover ADK-117

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#96**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

ADK-004 covers the Python path-safety case; the TypeScript half was missing. ADK-117 mirrors CSDK-012 including its coarse-signal caveat, stated in the explanation: it flags *any* filesystem write, not only unnormalized paths, because TS path-normalization analysis isn't wired yet. Confidence 0.5 to match.

**The callback gate doesn't close this, and that's the mitigation an ADK author would reach for.** ADK-102 and ADK-107 check that `before_tool_callback` is *present at all* — a callback that exists still has to inspect the path and know which root is allowed. Being wired up isn't the same as containing anything. That's why the `fix` puts the check **inside the tool**: it then holds for every agent in the tree that lists it, rather than only those whose callback was configured.

## What this PR does

1. Mirrors `google_adk/path_safety.yaml` into `testdata/rules-fixture/`.

Snippets use adk-js's options-object form (`new FunctionTool({ ..., execute })`) rather than the Python `FunctionTool(fn)` wrapper shape, which doesn't discover in TS — same fixture-shape note as #160.

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,28 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+
+	// ─── ADK-117: TS ADK FunctionTool writes to the filesystem ──────────────
+	{
+		name: "ADK-117 fires on filesystem write", ruleID: "ADK-117",
+		kind: models.KindADKFunctionTool, lang: models.LanguageTypeScript, wantFires: true,
+		src: "import { FunctionTool } from \"@google/adk\";\n" +
+			"import { writeFileSync } from \"node:fs\";\n" +
+			"const t = new FunctionTool({ name: \"save_note\", description: \"Save a note to disk for later retrieval.\", parameters: {}, execute: async ({ p, body }) => {\n" +
+			"  writeFileSync(p, body);\n" +
+			"  return \"saved\";\n" +
+			"} });\n",
+	},
+	{
+		name: "ADK-117 silent with no filesystem write", ruleID: "ADK-117",
+		kind: models.KindADKFunctionTool, lang: models.LanguageTypeScript, wantFires: false,
+		src: "import { FunctionTool } from \"@google/adk\";\n" +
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. This is the engine half done properly. The fixture copy of `google_adk/path_safety.yaml` is byte identical to the production half in trustabl/trustabl-rules#96, and putting both on a branch of the same name means the `rules-sync` job resolves the paired pack rather than `main`.

The two test cases are not vacuous, which is the part I check hardest. `parseTSTool` runs the real `DiscoverTSADKTools`, and both snippets carry the `@google/adk` import gate and the options-object `new FunctionTool({ ..., execute })` shape discovery recognizes, so neither case passes by failing to produce a tool. The fire case uses a bare `writeFileSync` callee, which is in the `writes_fs` set in `tsHandlerFacts`. The silent case is the useful one: `notes.set(...)` sets no fact, so the tool still satisfies `Applies` and reaches `Detect` and stays quiet on the predicate itself rather than short circuiting on an `applies_to` mismatch.

The blocking item is the rule ID, and I left the detail on trustabl/trustabl-rules#96: this is a tool-scoped rule sitting in the `1NN` block the README reserves for agent and subagent scope, and `ADK-014` is free. Both halves have to move together, so the fixture YAML and the two `ruleID` fields here need the same renumber.

Small nit while you are in the file: the diff also deletes a blank line at the tail of `policyRepoRuleCases`, which is unrelated here and adds needless conflict surface with the other open PRs appending to that slice, so please drop that hunk.

Happy to take this once the ID is renumbered alongside the rules half and the whitespace hunk is dropped.

PR #175, opened 2026-08-24
test(langchain): add LC-103 fixture coverage

jiaxinaspenlin-dotcom:
## Summary

Adds the engine-side fixture and test coverage for LC-103, which detects LangChain agents that wire `langchain_community`'s `Requests*` built-in HTTP tools.

## What changed

- Mirrored LC-103 into:
  - `testdata/rules-fixture/langchain/agent_safety.yaml`
- Added three agent-rule test cases in:
  - `internal/rules/policies_test.go`

Coverage includes:

- FIRE: `ReactAgent` with `RequestsGetTool`
- FIRE: `AgentExecutor` with `RequestsPostTool`
- SILENT: `ReactAgent` with `TavilySearchResults`

Every `AgentDef` sets `Language`.

No analyzer changes were required because the existing LangChain hosted-tool discovery already recognizes all five `Requests*` classes.

## Validation

- `go vet ./...` — pass
- `go test -race ./internal/rules/...` — pass
- `go test -race ./...` — 26 packages passed, 0 failures
- `go build -o /tmp/trustabl ./cmd/trustabl` — pass
- `check-rules-sync.sh` — in sync with production

The fixture is byte-identical to the corresponding rule definition in `trustabl-rules`.

## Paired changes

- trustabl/trustabl-rules#103 — LC-103 rule definition
- trustabl/trustabl-rulebook#79 — LC-103 rationale documentation

All coordinated changes use the shared branch:

`feat/langchain-requests-builtin-tool`

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -2505,6 +2505,25 @@ var policyAgentRuleCases = []policyAgentCase{
 			}}},
 		models.RepoInventory{}, false},
 
+	{"LC-103 fires when agent wires RequestsGetTool", "LC-103",
+		models.AgentDef{
+			SDK: models.SDKLangChain, Class: "ReactAgent", Language: models.LanguagePython,
+			HostedToolRefs: []models.HostedToolRef{{Class: "RequestsGetTool"}},
+		},
+		models.RepoInventory{}, true},
+	{"LC-103 fires when an AgentExecutor wires RequestsPostTool", "LC-103",
+		models.AgentDef{
+			SDK: models.SDKLangChain, Class: "AgentExecutor", Language: models.LanguagePython,
+			HostedToolRefs: []models.HostedToolRef{{Class: "RequestsPostTool"}},
+		},
+		models.RepoInventory{}, true},
+	{"LC-103 silent on a benign hosted tool", "LC-103",
+		models.AgentDef{
+			SDK: models.SDKLangChain, Class: "ReactAgent", Language: models.LanguagePython,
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @jiaxinaspenlin-dotcom, this is a clean engine half. I compared the `testdata/rules-fixture/langchain/agent_safety.yaml` hunk against the one in trustabl/trustabl-rules#103 and it matches line for line, so `rules-sync` stays green provided the two land together. The three cases follow the LC-101 pattern directly above them, set `Language`, and the two fire cases exercise different `applies_to` kinds (`ReactAgent` resolves to `langchain_agent`, `AgentExecutor` to `langchain_agent_executor`), which is more than the minimum the coverage guard asks for.

Two nits, neither blocking. `applies_to` also lists `langchain_state_graph` and nothing covers it, so mirroring LC-101's `StateGraph` case would close that out. The silent case uses `TavilySearchResults`, which is not in `LangChainHostedToolClasses` and so is never emitted as a `HostedToolRef` by a real scan; it does exercise the class filter, but an agent with no hosted refs would be the shape users actually hit.

Coordination note: LC-103 is also claimed by trustabl/trustabl#183 and trustabl/trustabl#186, so whichever lands first, the others will need to renumber and rebase. That is not a mark against this one.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-103`.

Closing trustabl/agent-reliability-rules#113 (a human-review approval gate, a different rule) and trustabl/agent-reliability-rules#108 (the same rule but matching only `RequestsGetTool`; this PR covers all five Requests classes).

sairenchristianbuerano (approved):

PR #176, opened 2026-08-24
test(rules): mirror and cover LC-024

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#104**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

## What this PR does

| case | expectation |
|---|---|

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,21 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "LC-024 fires on a generic tool name", ruleID: "LC-024", kind: models.KindLangChainTool, src: `
+def process(payload: str) -> str:
+    """Handle an incoming payload and return the result of handling it."""
+    return payload
+`, wantFires: true},
+	{name: "LC-024 silent on a verb-object tool name", ruleID: "LC-024", kind: models.KindLangChainTool, src: `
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy is byte identical to the rules half in trustabl/trustabl-rules#104, and the base `langchain/tool_definition.yaml` already matches production, so there is no drift hiding underneath the addition and `rules-sync` will be satisfied once the two land together.

The three cases are right, and the third is the one I would have asked for. `name_in` is an exact case-insensitive match rather than a substring, so `process_invoice_batch` is exactly the case that pins that, and because `parsePy` builds `ToolDef.Name` from the function name, all three exercise the predicate for real rather than passing vacuously. The fire case is named `process` (on the list) and the silent cases are not, so both halves of what `TestPolicyRules_AllRulesCovered` requires are genuinely covered rather than just present.

The stray blank line dropped at the end of `policyRepoRuleCases` is a gofmt tidy, no objection to it riding along.

The only thing still outstanding is on the rules side: LC-024 needs a rationale doc in `trustabl-rulebook` before it ships. That one is our job rather than yours, and I am happy to pair on it. I will take these two as a pair.

PR #177, opened 2026-08-24
test(rules): mirror and cover CREW-013

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#105**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

## What this PR does

1. Mirrors `crewai/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,21 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "CREW-013 fires on a generic tool name", ruleID: "CREW-013", kind: models.KindCrewAITool, src: `
+def process(payload: str) -> str:
+    """Handle an incoming payload and return the result of handling it."""
+    return payload
+`, wantFires: true},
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The mirror is exact: comparing the added block here against trustabl/trustabl-rules#105 line for line, the two are identical, which is what `check-rules-sync.sh` needs.

The three cases are right, and the third is the one worth having. `PredNameIn` lowercases and compares the whole name, so `process_invoice_batch` staying silent pins the semantics that would otherwise flood every pack sharing that list. `parsePy` takes `ToolDef.Name` from the function name, so all three cases exercise the predicate rather than passing vacuously.

I have left a change request on the rules half about CrewAI's positional decorator name, since `@tool("Fetch order status")` over `def process(...)` fires today on a name the model never sees. Whatever lands there has to be mirrored here, so I would rather hold this one until that settles. One unrelated nit while you are in there: the diff also drops a blank line at the end of `policyRepoRuleCases`, which is not part of this change and just adds noise to the review.

Happy to take it once the rules half is settled and mirrored back here.

PR #178, opened 2026-08-24
test(rules): mirror and cover AG2-019

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#106**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

## What this PR does

1. Mirrors `autogen/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,21 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "AG2-019 fires on a generic tool name", ruleID: "AG2-019", kind: models.KindAutoGenTool, src: `
+def process(payload: str) -> str:
+    """Handle an incoming payload and return the result of handling it."""
+    return payload
+`, wantFires: true},
+	{name: "AG2-019 silent on a verb-object tool name", ruleID: "AG2-019", kind: models.KindAutoGenTool, src: `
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy is byte identical to the YAML in trustabl/trustabl-rules#106, and both branches start from the same blob on `main`, so the sync check will be clean as long as the two land together.

The three cases are right, and the third is the one I would have asked for. `PredNameIn` lowercases and compares the whole name, so `process_invoice_batch` pins exact-match behavior rather than substring matching, which is exactly the regression that would flood every pack using this list. `parsePy` takes `Name` from the function name, so all three snippets exercise what you say they do, and none of them passes vacuously.

Two small things. The AG2-019 cases sit at the end of `policyRuleCases` after the CSDK-018 block, but that table is grouped by pack and the other AutoGen cases (AG2-007 through AG2-012) are much further up. Please move them next to those. The diff also drops a blank line at the tail of `policyRepoRuleCases`, which is unrelated to this change, so revert that hunk if it is easy.

Neither of those is blocking, and this is good to go alongside the rules half.

PR #179, opened 2026-08-24
test(rules): mirror and cover PYD-013

bradAGI:
> Engine half of a coordinated pair. Rules half: **trustabl/trustabl-rules#107**, on a branch of the **same name**, so the `rules-sync` job resolves the matching pack rather than `main`. Neither half should merge alone — `check-rules-sync.sh` fails if they do.

## What the pair adds

## What this PR does

1. Mirrors `pydantic_ai/tool_definition.yaml` into `testdata/rules-fixture/`.

| case | expectation |
|---|---|

## Verification

```

$ go vet ./internal/rules/
$ go test ./internal/rules/
```

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -1986,6 +1986,21 @@ def fetch_data(x: str) -> dict:
     return {}
 `,
 		toolConfig: nil, wantFires: false},
+	{name: "PYD-013 fires on a generic tool name", ruleID: "PYD-013", kind: models.KindPydanticAITool, src: `
+def process(payload: str) -> str:
+    """Handle an incoming payload and return the result of handling it."""
+    return payload
+`, wantFires: true},
+	{name: "PYD-013 silent on a verb-object tool name", ruleID: "PYD-013", kind: models.KindPydanticAITool, src: `
+def summarize_invoice(invoice_id: str) -> str:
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @bradAGI. The fixture copy of `PYD-013` is byte identical to the production YAML in trustabl/trustabl-rules#107, which is what `check-rules-sync.sh` is looking for, and the three cases in `policyRuleCases` all do real work. `parsePy` sets `ToolDef.Name` from the function name and `PredNameIn` is a whole-name comparison, so `process` fires, `summarize_invoice` stays silent, and `process_invoice_batch` genuinely pins the substring boundary rather than passing vacuously. Adding that third case was the right call, since it is the regression that would matter across every pack sharing this list.

Nothing blocking on this side. Two coordination notes. trustabl/trustabl#143 mirrors `PYD-010` and `PYD-011` into the same fixture file at the same offset and into the same `policies_test.go` hunk, so whichever of the two merges first leaves the other needing a rebase. And this half should land together with trustabl/trustabl-rules#107 rather than on its own.

I left a question on the rules half about the decorator `name=` override path, since `buildTool` takes the name from the function and ignores decorator kwargs. If we end up adding a guard for that, this file picks up a fourth case. The unrelated blank line dropped from `policyRepoRuleCases` is fine, just noting I saw it.

PR #180, opened 2026-08-24
fix(rules): make has_print_call fire on TypeScript tools

bradAGI:
## The bug

`PredHasPrintCall` walks for the tree-sitter node type `"call"` — which **does not exist in the TypeScript grammar**, where it's `call_expression`. The predicate is therefore structurally always false for a TS tool, so any `language: typescript` rule using `has_print_call` would load, validate, pass `trustabl rules validate`, and then never fire.

Nothing shipped hits this today because the observability rules (OAI-010, ADK-009) are python-only. I found it while drafting TS counterparts for the packs that have none: the rule validated cleanly and stayed silent on a fixture that plainly logs.

## The fix

Branch on `models.IsTSOrJS` and read a discovery-computed fact — exactly the shape `PredHasShellCall` and `PredHasWriteCall` already use for the same reason. **The Python path is untouched**, byte for byte.

`tsHandlerFacts` sets `prints_stdout` for:

| callee | why |
|---|---|
| `console.log`, `console.info`, `console.debug` | Node routes all three to **stdout** |
| `process.stdout.write` | the explicit form |

**Deliberately excluded: `console.warn`, `console.error`, `process.stderr.write`.** Node routes those to stderr, which is the correct destination for diagnostics — and on an MCP stdio server it's the *only* safe one, since stdout carries the JSON-RPC frames. Flagging them would report the remediation as the defect. There's a unit test asserting that exclusion specifically so it can't be "tidied up" into the same case list later.

## Why it's worth having

It unblocks TS observability rules across every pack. The sharpest one is MCP: in a TypeScript stdio server, `console.log` in a tool handler interleaves with the JSON-RPC frames and the client hits a parse error on a line that isn't JSON — a protocol bug, not a lost log line. The MCP TS SDK warns about exactly this. I'll follow up with that rule (and its rules-repo pair) once this lands.

## Verification

```
$ go vet ./...
$ go test ./...
(all packages ok)

$ go test ./internal/analysis/ -run TestTSHandlerFacts_PrintsStdout -v
--- PASS: TestTSHandlerFacts_PrintsStdout_ConsoleLogHits
--- PASS: TestTSHandlerFacts_PrintsStdout_ProcessStdoutWriteHits
--- PASS: TestTSHandlerFacts_PrintsStdout_ConsoleInfoAndDebugHit
--- PASS: TestTSHandlerFacts_PrintsStdout_StderrIsSilent
--- PASS: TestTSHandlerFacts_PrintsStdout_UnrelatedCalleeIsSilent
ok  	github.com/trustabl/trustabl/internal/analysis
```

`schema.yaml`'s `has_print_call` entry is updated to document the per-language behavior and the stderr exclusion.

No rules-repo pair needed: this adds no YAML key and changes no existing rule's behavior (every shipped `has_print_call` rule is python-only), so `check-rules-sync.sh` is unaffected and no `schema_version` bump is required. A pack that later ships a TS rule using it would want the bump then, to warn on older engines.

## Related

Two sibling gaps I hit in the same audit, not fixed here — happy to send either as its own PR:

- **`PredHasRaise` / `PredHasTryExcept`** match `raise_statement` / `try_statement`; TS throws parse as `throw_statement`, so TS error-contract rules are inert. This blocks a TS half for CSDK-005, OAI-008, ADK-005, MCP-006 and the four newer packs at once.
- **`ToolDef.Name` is empty for `vercel_ai_tool`** (asserted in `ts_vercel_tools_test.go` — the name lives in the `tools:` record key), so `name_in` / `name_has_prefix` can never match there. That's why the Vercel pack has no ambiguous-name or idempotency rule.


jhumel-code:
Thanks @bradAGI, good catch and the diagnosis holds up. `PredHasPrintCall` walks for the `call` node type, which only exists in the Python grammar, so the predicate was structurally always false for a TS tool. Both shipped users (`OAI-010` and `ADK-009`) are `language: python`, so no existing rule changes behavior, and I agree no rules-repo pair or `schema_version` bump is needed here. Branching on `models.IsTSOrJS` and reading a discovery fact is the same shape `PredHasShellCall`, `PredHasWriteCall` and `PredHasCodeExecCall` already use, and keeping `console.warn`, `console.error` and `process.stderr.write` out of the case list is right, as is the test locking that exclusion in.

One nit: the godoc above `PredHasPrintCall` still says only that it matches the bare `print` callee. `PredHasCodeExecCall` documents its TS branch in its comment, so this one should too.

One thing for the follow-up. `has_print_call` is a known key on older engines, so a pack shipping a TS rule that uses it will not be skipped by an older build, it will load and quietly return false. A `schema_version` bump buys the "rules newer than this build" warning, not a skip, so that limitation is worth stating in the rulebook doc for the MCP rule. `ts_mcp_proper.go` does route handlers through `tsHandlerFacts`, so that rule will have what it needs. Please send the `throw_statement` and Vercel `ToolDef.Name` gaps as their own PRs.

sairenchristianbuerano (approved):

PR #181, opened 2026-08-24
Hackathon/langchain hitl approval

dezsow:
Mirrors LC-112 into the engine rule fixture and adds focused fire/silent regression tests. Coverage includes missing approval prerequisites, middleware without a checkpointer, an empty middleware list, the structurally safe case, and agents without privileged execution tools.
This does not change LangChain discovery, predicate semantics, or evaluator behavior; it exercises existing machinery against the new rule.

### DIFF
--- a/internal/rules/lc112_policy_test.go
+++ b/internal/rules/lc112_policy_test.go
@@ -1,0 +1,101 @@
+package rules_test
+
+import (
+	"testing"
+
+	"github.com/trustabl/trustabl/internal/models"
+)
+
+func TestLC112HumanApprovalPrerequisites(t *testing.T) {
+	middleware := func(items ...models.Expr) *models.KwargTree {
+		return &models.KwargTree{
+			Value: &models.Expr{
+				Kind: models.ExprList,
+				List: items,
+			},
+		}
+	}
+	checkpointer := &models.KwargTree{
+		Value: &models.Expr{Kind: models.ExprCall, Text: "InMemorySaver()"},
+	}
+
+	cases := []struct {
+		name      string
+		agent     models.AgentDef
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks for splitting this cleanly across the two repos, @dezsow. The fixture addition matches the YAML in trustabl/trustabl-rules#110 byte for byte, so `rules-sync` stays happy, and you kept to existing machinery: `agent_class`, `agent_uses_hosted_tool_class`, `agent_kwarg_list_empty` and `agent_kwarg_missing` all exist on `main` at schema 14, so no version bump is warranted. The case selection is good too, particularly the empty-`middleware`-with-`checkpointer`-present case, which is the one that actually exercises the `any` branch rather than passing by accident.

One blocking item. `TestPolicyRules_AllRulesCovered` builds its coverage set only from `policyRuleCases`, `policyAgentRuleCases`, `policyRepoRuleCases`, `policySubagentRuleCases` and `policySkillRuleCases` in `internal/rules/policies_test.go`. A table in a standalone `lc112_policy_test.go` is invisible to it, so once LC-112 is in the fixture the guard will report it as having no coverage at all and fail. Please add a fire and a silent `policyAgentCase` for LC-112 to `policyAgentRuleCases` (keep the dedicated file as extra depth if you like), with `Language` set on every `AgentDef` so `TestFixtureAgentsHaveLanguage` stays green.

I also left a severity question on the rules PR that would change this YAML, so the two want to land together. LC-112 will need a rationale doc in `trustabl-rulebook` before it ships; that part is ours and we are happy to pair on it.

Happy to take this once the coverage entries are in `policies_test.go` and the rules-side question is settled.

PR #182, opened 2026-08-24
feat(rules): VarName name match; cover VAI-010 and VAI-016

joeymussalli:
## Summary
- `name_has_prefix` / `name_in` fall back to `VarName` when `ToolDef.Name` is empty (required for Vercel AI tools).
- Mirror + fire/silent coverage for **VAI-010** (mutating tool idempotency) and **VAI-016** (ambiguous binding name).

## Why
Vercel discovery leaves `Name` blank and puts the binding identifier in `VarName`. Without the fallback, every name-based Vercel rule is a no-op. The rulebook's Vercel tool-definition doc even called name-based rules a deliberate coverage gap — this closes it.

## Paired PRs
- Rules VAI-010: trustabl-rules#109
- Rules VAI-016: (new)
- Rulebook VAI-010: trustabl-rulebook#94
- Rulebook VAI-016: (new)

## Test plan
- [ ] `go test ./internal/rules/ -run 'NameHasPrefix|VAI-010|VAI-016|AllRulesCovered'`
- [ ] Full `go test ./...` in CI

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -362,6 +362,18 @@ import { z } from "zod";
 export const t = tool({ description: "x", inputSchema: z.object({ city: z.string() }), execute: async ({ city }) => city });
 `},
 
+	// VAI-016: ambiguous binding name (VarName) — Name is empty for Vercel tools.
+	{name: "VAI-016 fires on process binding", ruleID: "VAI-016", kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: true, src: `
+import { tool } from "ai";
+import { z } from "zod";
+export const process = tool({ description: "x", inputSchema: z.object({ q: z.string() }), execute: async ({ q }) => q });
+`},
+	{name: "VAI-016 silent on descriptive binding", ruleID: "VAI-016", kind: models.KindVercelAITool, lang: models.LanguageTypeScript, wantFires: false, src: `
+import { tool } from "ai";
+import { z } from "zod";
+export const fetchWeather = tool({ description: "x", inputSchema: z.object({ city: z.string() }), execute: async ({ city }) => city });
+`},
+
 	// ─── CrewAI tool rules (CREW-*) ─────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


jhumel-code:
Thanks @joeymussalli, this is well diagnosed. I read `internal/analysis/ts_vercel_tools.go` on `main` and you are right that Vercel discovery leaves `ToolDef.Name` empty and captures the binding in `VarName`, so every name based rule really was inert there. Both production halves exist (`trustabl/trustabl-rules#109` and `trustabl/trustabl-rules#112`), the VAI-010 file matches the rules repo copy exactly, and the severity and confidence values sit right alongside `CSDK-016` and `OAI-007`.

Two things I want changed first. The fallback in `toolNameForMatch` changes `name_in` and `name_has_prefix` for every rule in every pack, not just Vercel: `CSDK-016` and `OAI-019` are TypeScript `name_has_prefix` rules that will now match the binding identifier whenever a `name:` literal is absent. If that widening is intended, please pin it with a case, and either way `internal/rules/schema.yaml` still documents both predicates as matching the tool name and needs updating in the same change. Note also that because no new predicate key was introduced, an engine build without this fallback will silently never fire VAI-010 or VAI-016 and will record no skip.

Second, VAI-010 should use `param_name_matches` the way `CSDK-016` does instead of `has_body_text`. Vercel discovery populates `ParamNames` from the Zod schema, so the structured predicate works here, whereas `has_body_text` is case sensitive across the whole `tool(...)` span, so it will not suppress on `IdempotencyKey` and will suppress on the token appearing in a comment.

Two smaller notes. `trustabl/trustabl-rules#112` drops the trailing newline on `vercel_ai/tool_definition.yaml` while this fixture copy keeps it, which `scripts/check-rules-sync.sh` reports as content drift, so please fix that on the rules side. And `directAssignmentName` only resolves `const x = tool(...)`, so a tool declared inline in the `tools:` record has an empty `VarName` and both rules stay silent on it; worth saying so in the explanation. Both rules also need rationale docs in `trustabl-rulebook` before they ship, and we can pair on that or take it ourselves.

Happy to merge this once the predicate widening is settled and documented and VAI-010 moves to `param_name_matches`.

sairenchristianbuerano (approved):

PR #183, opened 2026-08-24
test(rules): add LC-101 PythonREPL regression case, LC-103/LC-104 coverage

ShonenAsh:
PR submitted at Hult x Cursor hackathon. AI use was encouraged. I normally do not use Agents, but the 3 hr hackathon required me to.
-Shonen

Mirrors the langchain/agent_safety.yaml change from trustabl-rules into testdata/rules-fixture/ and adds the fire/silent cases TestPolicyRules_AllRulesCovered requires for the two new rule IDs.

The LC-101 case is a regression guard, not a formality: dropping the PythonREPL entry from the fixture makes it fail with "rule LC-101: fired=false, want true", which is the false negative the rules-repo change closes.

The LC-103 and LC-104 silent cases cross over on purpose -- LC-103 must stay quiet on a write verb and LC-104 on the read verb. A silent case that merely omits the hosted-tool edge would pass even if both rules matched the same class list, so it would not pin the read/write split the rules describe.

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -2484,6 +2484,15 @@ var policyAgentRuleCases = []policyAgentCase{
 	{"LC-101 silent on a StateGraph with no dangerous builtin", "LC-101",
 		models.AgentDef{SDK: models.SDKLangChain, Class: "StateGraph", Language: models.LanguagePython},
 		models.RepoInventory{}, false},
+	// The bare PythonREPL utility is in LangChainHostedToolClasses as a
+	// code-execution class, so `tools=[PythonREPL()]` is discovered as a hosted
+	// edge — but LC-101 did not list it, so an RCE-class wiring scanned clean.
+	{"LC-101 fires when agent wires the bare PythonREPL utility", "LC-101",
+		models.AgentDef{
+			SDK: models.SDKLangChain, Class: "ReactAgent", Language: models.LanguagePython,
+			HostedToolRefs: []models.HostedToolRef{{Class: "PythonREPL"}},
+		},
+		models.RepoInventory{}, true},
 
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @ShonenAsh. The fixture change here is byte identical to trustabl/trustabl-rules#108, which is exactly what rules-sync needs, and the test cases are above the usual bar. The `LC-101` case is a real regression guard rather than a formality, and the crossed silent cases (`RequestsPostTool` under `LC-103`, `RequestsGetTool` under `LC-104`) pin the read/write split itself, so neither rule can quietly widen into the other's verbs without the suite noticing. Walking the five cases against the predicate, `PredAgentUsesHostedToolClass` matches on `HostedToolRef.Class` without requiring resolution, and `Class: "ReactAgent"` with `SDKLangChain` satisfies the `langchain_agent` token, so each case exercises the path it claims to. Nothing here fires vacuously.

The only thing outstanding is coordination, not code. `LC-103` is also claimed by trustabl/trustabl#175 and trustabl/trustabl#186 (rules halves trustabl/trustabl-rules#103 and trustabl/trustabl-rules#113); whichever pair lands first, the others will need to renumber and rebase, and sequencing that is our job rather than yours. `LC-103` and `LC-104` will also need rationale docs in `trustabl-rulebook` before they ship, which we can take on or pair with you on.

I am happy to take this together with trustabl/trustabl-rules#108 once the numbering shakes out.

sairenchristianbuerano:
**Closing.** `LC-103` goes to trustabl/agent-reliability-rules#103. Both PRs write the same rule, but #103 matches all five Requests classes (Get/Post/Put/Patch/Delete) while this one matches `RequestsGetTool` only, so #103 is the superset.

Your `LC-104` and the `LC-101` PythonREPL work are not duplicated anywhere and close with this PR. They would be worth resubmitting separately.

PR #184, opened 2026-08-24
docs(skill): list every --detectors category the CLI accepts

Paramjeet-singh-neu:
## Summary
\`skills/trustabl-scan/SKILL.md\` still told the CLI-fallback that \`--detectors\` is \`claude_sdk|openai_sdk|google_adk|openshell\`. The flag accepts every \`models.AllCategories\` value, including \`mcp\`, \`langchain\`, \`crewai\`, \`pydantic_ai\`, \`vercel_ai\`, \`autogen\`, and \`claude_skill\`.

One comment. No rule IDs. Independent of the CLI-help PRs.

## Test plan
- [ ] \`trustabl scan --help\` \`--detectors\` line matches the updated comment

### DIFF
--- a/skills/trustabl-scan/SKILL.md
+++ b/skills/trustabl-scan/SKILL.md
@@ -74,7 +74,7 @@ on `PATH`:
 ```bash
 "$TRUSTABL_BIN" scan . --format json          # JSON, same shape the tool returns
 "$TRUSTABL_BIN" scan . --strict               # exit 1 on low+ findings (info/META never fail)
-"$TRUSTABL_BIN" scan . --detectors claude_sdk # narrow to one SDK: claude_sdk|openai_sdk|google_adk|openshell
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu, this is a real fix and the claim holds up. On `main`, `models.AllCategories` carries all eleven values (`claude_sdk`, `openai_sdk`, `openshell`, `google_adk`, `mcp`, `langchain`, `crewai`, `pydantic_ai`, `vercel_ai`, `autogen`, `claude_skill`), and `parseCategories` validates against exactly that list via `ValidCategory`, so every one of them is accepted. That comment in `skills/trustabl-scan/SKILL.md` was the only spot left in the docs still enumerating just the original four, and nothing else in that file needs to move with it.

Two optional notes, neither blocking. The `--help` string is generated by `categoryList()`, which prints in `models.AllCategories` order, so `openshell` comes third there rather than fourth; matching that order would turn your test-plan step into a straight string comparison. Separately, `README.md` already notes that `--detectors openshell` is accepted but currently selects zero rules because that pack is closed source, so someone picking it off this list gets an empty run. Neither is worth another round trip.

Happy to merge this as is.

sairenchristianbuerano (approved):

PR #185, opened 2026-08-24
feat(pydantic): detect FileUrl and WebFetchTool force_download (PYD-104)

Paramjeet-singh-neu:
## Summary
- Close the documented v1 gap for **PYD-104**: Pydantic AI `FileUrl` / `ImageUrl` / `AudioUrl` / `VideoUrl` / `DocumentUrl` and `WebFetchTool` with `force_download=True` or `"allow-local"` fetch on the **agent host** (SSRF; CVE-2026-46678, CVE-2026-25580). Default `False` stays silent.
- New predicate `agent_file_url_force_download`, FileUrl-family discovery stamped on `AgentDef`, and `NativeTool(...)` inner-kwarg unwrap so `WebFetchTool(force_download=...)` is visible. Rule schema **15**.
- Companions: rules https://github.com/trustabl/trustabl-rules/pull/111 · rulebook https://github.com/trustabl/trustabl-rulebook/pull/99

## Test plan
- [x] `go test ./internal/rules/ ./internal/analysis/` (fire/silent cases + FileUrl / NativeTool unwrap discovery)
- [x] Reviewer: scan a Pydantic AI file with `DocumentUrl(..., force_download=True)` and one with default/`False`


jhumel-code:
Thanks @Paramjeet-singh-neu, this is a well built change. The schema bump is done properly: `schema.go`, `predicates.go`, `evaluator.go` and `schema.yaml` all move together, `SupportedSchemaVersion` goes to 15 alongside both manifests, and the fixture YAML is byte identical to trustabl/trustabl-rules#111. `agent_file_url_force_download` is registered in `predicatesByScope` and `setPredicateNames`, so the drift tests are satisfied, and unwrapping `NativeTool(...)` through the new `Expr.CallArgs` is the right place for that work: the predicate reads discovery output instead of re-walking the AST. `ARCHITECTURE.md` and `COVERAGE.md` are updated in the same change, including the PYD-103 `WebSearchTool` correction.

Two asks. The third `any` branch, `WebFetchTool` with `force_download: "allow-local"`, has no case in `policies_test.go`, and the coverage guard only requires one fire and one silent per rule, so that arm ships unexercised. Second, the `force_download=True` hosted-tool case builds the kwarg as `ExprNameRef`, while `exprFromNode` stamps Python `True` as `ExprLiteralBool`. It passes either way (only string literals get quote stripped), but it encodes a shape the parser never produces.

I have also asked for a confidence change on trustabl/trustabl-rules#111; please mirror it here so the fixture stays in sync. PYD-104 needs a rationale doc in `trustabl-rulebook` before it ships, which we can take or pair on.

Happy to take this once those are in.

sairenchristianbuerano (approved):

PR #186, opened 2026-08-24
test(langchain): add LC-103 fixture and regression cases

Mantitah:
Fork of the Trustabl core scanner, updated with the mirrored LC-103 rule fixture and regression tests covering unsafe and human-review-configured LangChain agents.

### DIFF
--- a/internal/rules/lc103_test.go
+++ b/internal/rules/lc103_test.go
@@ -1,0 +1,117 @@
+package rules_test
+
+import (
+	"testing"
+
+	"github.com/trustabl/trustabl/internal/models"
+)
+
+func TestLC103LangChainHumanReviewGate(t *testing.T) {
+	d := loadAgentRule(t, "LC-103")
+
+	cases := []struct {
+		name      string
+		agent     models.AgentDef
+		wantFires bool
+	}{
+		{
+			name: "fires when CreateAgent wires PythonREPLTool without HITL prerequisites",
+			agent: models.AgentDef{
+				SDK:            models.SDKLangChain,
+				Class:          "CreateAgent",
+				Language:       models.LanguagePython,
+				HostedToolRefs: []models.HostedToolRef{{Class: "PythonREPLTool"}},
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @Mantitah, this is a clean mirror. The fixture copy and the YAML in trustabl/trustabl-rules#113 are the same blob, which is exactly what `rules-sync` wants. Every predicate the rule uses (`agent_class`, `agent_uses_hosted_tool_class`, `agent_kwarg_list_empty`, `agent_kwarg_missing`) already exists on `main` at schema 14, `langchain_agent` is a valid `applies_to` at agent scope, and `CreateAgent` is the normalized class name LangChain discovery actually produces for `create_agent`. The five cases exercise both arms of the `any`, which I appreciate.

One blocking item. `TestPolicyRules_AllRulesCovered` builds its coverage set only from the tables in `internal/rules/policies_test.go` (`policyAgentRuleCases` for agent scope). Your cases live in a separate `internal/rules/lc103_test.go` with a local table, so the guard will not see them and will fail with LC-103 listed under "rules with no policy_test coverage at all". Please move the fire and silent cases into `policyAgentRuleCases`.

Two coordination notes. Both halves are on your fork's `main`, so the `rules-sync` job's same-branch lookup falls back to upstream `trustabl-rules` main and this PR stays red until the rules half merges. And LC-103 is also claimed by trustabl/trustabl#175 and trustabl/trustabl#183 for a different rule, so whichever lands first, the others renumber and rebase.

Happy to take this once the cases move into `policyAgentRuleCases` and the severity question on trustabl/trustabl-rules#113 is settled.

sairenchristianbuerano:
**Closing.** `LC-103` goes to trustabl/agent-reliability-rules#103, which uses it for the raw HTTP Requests built-in tool rule. Your resumable human-review approval gate is a different rule on the same id and is not shipping as a result. `LC-105` onward is free if it comes back.

PR #187, opened 2026-08-24
feat(vercel_ai): TS has_print_call + VAI-017 fixture

joeymussalli:
## Summary
- Teach `PredHasPrintCall` to read the TypeScript `prints_stdout` discovery fact (`console.log` / `console.info` / `console.debug` / `process.stdout.write`; stderr callees stay silent), mirroring `has_write_call` / `has_shell_call`.
- Add unit coverage for the fact and fire/silent cases for **VAI-017** (Vercel AI tool stdout diagnostics).
- Ship the matching rules-fixture so engine CI can evaluate the new rule.

## Overlap
Same `prints_stdout` / `PredHasPrintCall` TS branch as #180. If that lands first, I will rebase and drop the overlapping hunks, keeping only the VAI-017 fixture + policy tests. #180 telegraphed MCP follow-up; this claims the Vercel observability rule.

## Companion PRs
- Rules: https://github.com/trustabl/trustabl-rules/pull/114
- Rulebook: https://github.com/trustabl/trustabl-rulebook/pull/102

## Test plan
- [ ] `go test ./internal/analysis/ -run PrintsStdout`
- [ ] `go test ./internal/rules/ -run 'VAI-017|OAI-010'`

### DIFF
--- a/internal/analysis/ts_handler_facts.go
+++ b/internal/analysis/ts_handler_facts.go
@@ -64,6 +64,13 @@ func tsHandlerFacts(handler *sitter.Node, src []byte) map[string]string {
 			"fs.appendFileSync", "fs.createWriteStream",
 			"fsPromises.writeFile", "fsPromises.appendFile":
 			out["writes_fs"] = "true"
+		case "console.log", "console.info", "console.debug", "process.stdout.write":
+			// STDOUT ONLY. console.warn/console.error and process.stderr.write
+			// are deliberately absent: Node routes them to stderr, which is the
+			// correct destination for diagnostics and — on an MCP stdio server —
+			// the only safe one, since stdout carries the JSON-RPC frames.
+			// Flagging them would report the remediation as the defect.
+			out["prints_stdout"] = "true"
 		case "eval":
 			// Bare `eval` callee only — callee text for `retrieval(x)` is
 			// "retrieval", so this exact-match eliminates the false-positive.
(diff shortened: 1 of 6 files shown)


jhumel-code:
Thanks @joeymussalli, this is clean work. The `PredHasPrintCall` TypeScript branch is exactly right: the Python walk matches the `call` node type and the TS grammar uses `call_expression`, so the predicate really was structurally dead for TS, and mirroring `PredHasWriteCall` keeps the Python path untouched. Limiting the callee list to stdout and leaving `console.warn`, `console.error` and `process.stderr.write` silent is the correct call, and the silent cases cover it properly. `parseTSTool` runs real Vercel discovery, so the VAI-017 fire case exercises the fact end to end rather than passing vacuously. Both existing `has_print_call` rules are `language: python`, so nothing already shipped changes behavior.

Two things before I take it. Docs are part of the change in this repo: `ARCHITECTURE.md` enumerates the `tsHandlerFacts` keys and the list of TS-aware predicates (both need `prints_stdout` and `PredHasPrintCall`), and `COVERAGE.md`'s Vercel row lists the shipped VAI rule IDs. And the predicate and facts hunks duplicate trustabl/trustabl#180 almost verbatim. Thanks for calling that out in the description; whichever lands first, the other should rebase down to only its own additions.

Please also keep the fixture byte-identical to trustabl/trustabl-rules#114, including the explanation rewrite I asked for over there.

Happy to take this once those are sorted.

sairenchristianbuerano (approved):

PR #188, opened 2026-08-24
fix(scanner): count all source languages in recon summary

Paramjeet-singh-neu:
## What

The recon-phase summary line pairs a file count with the **full** detected-languages label, but only counted Python files:

```go
// internal/scanner/scanner.go
rep.EndPhase(fmt.Sprintf("%d files · %s",
    len(profile.Manifest.PythonFiles), languagesLabel(profile.Languages)))
```

So a TypeScript-only repo prints `0 files · typescript`, and any multi-language repo undercounts. The inventory `SetTotal` on the very next lines already summed **all seven** source-file buckets (Python, TypeScript, JavaScript, Go, C#, PHP, Rust), so this label was the odd one out.

## Change

- Add `ScanManifest.SourceFileCount()` — sums every language the inventory step parses; data files (YAML/JSON/Markdown) are excluded since they are not AST-parsed.
- Use it in both the recon summary label (the fix) and the inventory `SetTotal` (removes the duplicated seven-term sum).
- Add a unit test covering the empty case, the all-buckets case (data files excluded), and a TypeScript-only regression.

## Impact

This is **stderr-only progress output** (`internal/progress`). Per the determinism contract, it never touches the byte-stable report or `ScanID`, so no report changes and no doc updates are required.

## Testing

- `go build ./...`
- `go vet ./internal/models/ ./internal/scanner/`
- `go test ./...` — all green

### DIFF
--- a/internal/models/models.go
+++ b/internal/models/models.go
@@ -293,6 +293,16 @@ type ScanManifest struct {
 	Components             []AgentComponent `json:"components,omitempty"`
 }
 
+// SourceFileCount returns the number of source files across every language the
+// inventory step parses (Python, TypeScript, JavaScript, Go, C#, PHP, Rust).
+// This is the count that pairs with the detected-languages label — using a
+// single language's bucket would undercount any multi-language repo. Data files
+// (YAML/JSON/Markdown) are intentionally excluded: they are not AST-parsed.
+func (m ScanManifest) SourceFileCount() int {
+	return len(m.PythonFiles) + len(m.TypeScriptFiles) + len(m.JavaScriptFiles) +
+		len(m.GoFiles) + len(m.CSharpFiles) + len(m.PHPFiles) + len(m.RustFiles)
+}
+
 // SDK identifies a tool/agent SDK we know about.
 type SDK string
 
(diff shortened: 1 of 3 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu, good catch and a clean fix. You are right that the recon summary was the odd one out: `languagesFromManifest` in `internal/ingestion/normalizer.go` derives the label from exactly the same seven source buckets, so pairing that label with `len(PythonFiles)` was guaranteed to disagree on any repo that is not Python-first. Hanging `SourceFileCount()` off `ScanManifest` right next to the buckets is the right home for it, excluding the YAML/JSON/Markdown lists is correct since those are never AST-parsed, and the test covers the empty case, the all-buckets case, and the TypeScript-only regression rather than just restating the sum. Your read on determinism holds up too: `EndPhase` and `SetTotal` are progress-reporter calls, so nothing in the report or `ScanID` moves.

Two small things, neither blocking. There is a third copy of the same seven-term sum further down `internal/scanner/scanner.go`, where `filesAttempted` is built for `Coverage.FilesSkipped`; swapping the helper in there is byte-identical and would leave one definition instead of two. And `SourceFileCount()` and `languagesFromManifest` still enumerate the bucket list separately, so an eighth language means remembering both. Happy to take this as it stands.

sairenchristianbuerano (approved):

PR #189, opened 2026-08-24
docs(rules): correct RuleDef language-default comment

Paramjeet-singh-neu:
## What

The `RuleDef` doc comment claims the language default applies unconditionally:

```go
// Language defaults to "python" if absent in YAML — the loader fills it in.
```

But the loader only fills it in for **tool** and **agent** scope:

```go
// internal/rules/loader.go
if pf.Rules[i].Language == "" &&
    (pf.Rules[i].Scope == models.ScopeTool || pf.Rules[i].Scope == models.ScopeAgent) {
    pf.Rules[i].Language = models.LanguagePython
}
```

Subagent, skill, and repo rules do not gate on language (markdown frontmatter / inventory scopes), so an omitted language stays empty for them — which matches the loader's own comment and `CLAUDE.md`.

## Change

Doc-only: qualify the comment so it matches the code. No behavior change.

### DIFF
--- a/internal/rules/schema.go
+++ b/internal/rules/schema.go
@@ -18,7 +18,9 @@ type PolicyMeta struct {
 
 // RuleDef is one rule entry inside a policy file.
 // Category is not in YAML — the loader copies it from PolicyMeta.Category.
-// Language defaults to "python" if absent in YAML — the loader fills it in.
+// Language defaults to "python" for tool/agent scope if absent in YAML — the
+// loader fills it in. Subagent, skill, and repo rules do not gate on language,
+// so an omitted language stays empty for those scopes.
 type RuleDef struct {
 	ID          string                  `yaml:"id"`
 	Title       string                  `yaml:"title"`


jhumel-code:
Thanks @Paramjeet-singh-neu, this is a real defect and the diagnosis is right. `loadPolicies` only fills in `models.LanguagePython` when the scope is `tool` or `agent`, so the unqualified comment on `RuleDef` was misleading for the other three scopes. I like that you kept it doc-only.

One correction before I take it. The new second sentence says subagent, skill, and repo rules do not gate on language. That holds for subagent and skill (`subagentRuleDetector.Applies` and `skillRuleDetector.Applies` have no language gate at all), but `repoRuleDetector.Applies` does gate: when `d.rule.Language != ""` it walks `RepoProfile.Languages` and returns false if none matches, with an inline comment explaining why (a `language: python` repo rule would otherwise fire on a TS-only repo). Please narrow the claim to what the loader actually does, roughly: for subagent, skill, and repo scope an omitted language stays empty, and a repo rule that does declare one is still gated by it at detection time.

Two optional extras while you are in here, since both carry the same stale claim: `internal/rules/schema.yaml` around line 57 still reads `OPTIONAL (default: python)` with no scope qualification, and the loader's own comment above the default omits skill scope.

Happy to take this once the repo-scope clause is tightened.

sairenchristianbuerano (approved):

PR #190, opened 2026-08-24
fix(forge): reject a stamp whose sdks: field lacks its prefix

Paramjeet-singh-neu:
## What

`ParseStamp` guards the `rules:` and `schema:` fields by checking that `TrimPrefix` actually stripped the expected prefix:

```go
rulesVal := strings.TrimPrefix(parts[1], "rules: ")
if rulesVal == parts[1] { // prefix not found
    return Stamp{}, false
}
```

The `sdks:` field skipped that check:

```go
sdksStr := strings.TrimPrefix(parts[3], "sdks: ")
var sdks []string
for _, s := range strings.Split(sdksStr, ", ") { ... }
```

So a malformed 4th segment (missing the `sdks: ` prefix) was silently absorbed into the SDK list rather than being rejected — contradicting the doc contract: *"Returns `(Stamp{}, false)` when … the comment is malformed."*

## Change

- Add the same prefix guard used by the sibling fields.
- Add a regression test (`TestParseStamp_MalformedSDKsPrefix`).

`go test ./...` passes.


jhumel-code:
Thanks @Paramjeet-singh-neu, good catch and the fix is exactly the shape I want. The `sdks:` field was the only one of the four missing the `TrimPrefix` sentinel check, so a 4th segment like `claude_sdk, mcp` parsed clean and got absorbed into `Stamp.SDKs`, which contradicts the doc comment on `ParseStamp`. The new guard mirrors the `rules:` and `schema:` ones line for line.

I checked the case I was most worried about, the zero-SDK stamp. `Line()` emits `sdks: ` with a trailing space, so `parts[3]` is `"sdks: "`, `TrimPrefix` returns empty, and the guard does not trip, which leaves `TestStamp_Line_NoSDKs` and the round trip intact. The new test is non-vacuous too: its input clears the length, `rules:` and `schema:` checks and only fails at the new guard, so it would have returned true before this change.

The only caller is `runForgeCheck`, which turns `!ok` into exit 1 plus a regenerate hint, and that is the right outcome for a malformed stamp.

One optional nit: the new case would sit fine as another entry in the existing `TestParseStamp_NoStamp` table instead of its own function. Not holding the PR on that. Happy to take this as is.

sairenchristianbuerano (approved):

PR #191, opened 2026-08-24
test(langchain): cover missing network timeout rule

pjsk02:
## Summary

Adds the engine fixture and regression coverage for the new LangChain `LC-007` network-timeout rule.

The canonical rule uses Trustabl's existing `call_without_kwarg` predicate, so no production engine changes are required.

### Fire cases

- `requests.get(...)` without timeout
- `requests.post(...)` without timeout
- `requests.get(..., timeout=None)`
- `requests.Session().get(...)` without timeout

### Silent cases

- `requests.get(..., timeout=10)`
- bare `httpx.get(...)`
- LangChain tool with no network call

The HTTPX silent case deliberately guards against treating HTTP clients with different timeout defaults as equivalent to Requests.

## Validation

Passed:

- targeted policy-rule tests
- all 7 LC-007 regression cases
- `TestPolicyRules_AllRulesCovered`
- `go vet ./...`
- `go build ./cmd/trustabl`
- canonical/fixture byte-sync
- `git diff --check`

`go test ./...` has an existing environment-dependent failure under `internal/enrichment` when the expected Python runtime is unavailable. The same failure was independently reproduced from a clean upstream worktree without LC-007; the LC-007 policy tests pass.

## Companion PRs

- Rule definition: https://github.com/trustabl/trustabl-rules/pull/115
- Rulebook rationale: https://github.com/trustabl/trustabl-rulebook/pull/107

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -230,6 +230,50 @@ def f(x: str) -> str:
     return x
 `, toolConfig: map[string]string{}, wantFires: false},
 
+	// ─── LC-007 LangChain tool network call without timeout ──────────────────
+	{name: "LC-007 fires on requests.get without timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def fetch(q: str) -> str:
+    """Fetch."""
+    import requests
+    return requests.get("https://api.example.com/data").text
+`, wantFires: true},
+	{name: "LC-007 fires on requests.post without timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def create(q: str) -> str:
+    """Create."""
+    import requests
+    return requests.post("https://api.example.com/data", json={"q": q}).text
+`, wantFires: true},
+	{name: "LC-007 silent with timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def fetch(q: str) -> str:
+    """Fetch."""
+    import requests
+    return requests.get("https://api.example.com/data", timeout=10).text
+`, wantFires: false},
+	{name: "LC-007 fires on timeout=None", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+import requests
+def fetch(url: str) -> str:
+    """Fetch."""
+    return requests.get(url, timeout=None).text
+`, wantFires: true},
+	{name: "LC-007 silent on httpx.get without timeout", ruleID: "LC-007", kind: models.KindLangChainTool, src: `
+def fetch(q: str) -> str:
+    """Fetch."""
+    import httpx
+    return httpx.get("https://api.example.com/data").text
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @pjsk02. The fixture half is exactly right: `testdata/rules-fixture/langchain/network.yaml` is byte identical to the YAML in trustabl/trustabl-rules#115, and no schema or manifest change is correct here since `call_without_kwarg` already exists on main.

The cases are real rather than vacuous. Walking each against the predicate, the plain `requests.get` and `requests.post` snippets have no `timeout` kwarg so it fires, `timeout=None` fires on the literal-None branch, `timeout=10` is present and not None so it stays silent, `httpx.get` canonicalizes to `httpx.get` which is not in the callee list so it stays silent, and the echo tool has no matching call at all. Four fire cases and three silent, so the coverage guard is satisfied. One naming note: the session case fires because `ResolveClientAliases` canonicalizes `s.get` to `requests.get`, not through the `requests.Session.get` entry, so what it actually proves is alias resolution rather than that entry.

I have asked on trustabl/trustabl-rules#115 for the callee list to be widened to match the sibling network packs, urllib and aiohttp in particular. That changes the YAML here too, and the httpx silent case flips to a fire case if we also add httpx, so let us land this alongside the rules change rather than ahead of it. Note as well that `LC-007` and this same fixture file are claimed by trustabl/trustabl#127 and trustabl/trustabl-rules#52, so whichever lands first the other will need to renumber and rebase.

Happy to take this once the mirrored YAML matches the updated rule.

sairenchristianbuerano:
**Closing in favour of trustabl/agent-reliability-rules#52 + trustabl/agent-reliability-analyzer#127** for `LC-007`.

@jhumel-code asked here for the callee list to be widened to match the sibling network packs, urllib and aiohttp in particular. #52's list is already **byte-identical to `pydantic_ai/network.yaml` on main** — requests, httpx, `urllib.request.urlopen`, bare `urlopen` and the aiohttp ClientSession pair. Making the requested change here would have reproduced that file exactly, so we are taking the one that already has it.

Your fixture work was verified correct and byte-identical to the rules YAML; the review above confirms that.

sairenchristianbuerano:
**Correction to my previous comment — this PR should merge, not close.**

trustabl/agent-reliability-rules#115 was merged while that comment was being written, so `LC-007` now ships from it. That makes this fixture required rather than redundant: `langchain/network.yaml` is on rules `main` with no mirror here, so `rules-sync` fails until this lands.

Verified just now: the fixture on this branch is **byte-identical** to `langchain/network.yaml` as merged on rules `main`. No rebase needed — `mergeable: true`.

Apologies for the whiplash. Please merge this one.

sairenchristianbuerano (approved):

PR #192, opened 2026-08-24
feat(rules): add CSKILL-087 fixture and fire/silent cases

kureen-cyber:
## Summary
- Mirrors **CSKILL-087** into `testdata/rules-fixture/claude_skill/skill_quality_text.yaml`.
- Adds fire + silent cases in `policySkillRuleCases` so `TestPolicyRules_AllRulesCovered` stays green.

Paired with https://github.com/trustabl/trustabl-rules/pull/116 on the same branch name (`feat/cskill-087-placeholder-description`) so `rules-sync` can resolve the matching pack. That branch currently lives on the `kureen-cyber` fork; until it exists on `trustabl/trustabl-rules`, this job falls back to `main` and will report fixture drift. Merge or copy the rules branch first.

## Test plan
- [ ] `go test ./internal/rules/` — CSKILL-087 fire + silent pass
- [ ] `TestPolicyRules_AllRulesCovered` green
- [ ] `scripts/check-rules-sync.sh` against the matching rules branch

### DIFF
--- a/internal/rules/policies_test.go
+++ b/internal/rules/policies_test.go
@@ -2461,6 +2461,13 @@ var policySkillRuleCases = []policySkillCase{
 		models.SkillDef{Name: "helper",
 			Location: models.Location{FilePath: ".claude/skills/helper/SKILL.md"},
 			Body:     "# Helper\n\nSummarises the current git diff."}, models.RepoInventory{}, false},
+
+	{"CSKILL-087 fires when the description is a placeholder", "CSKILL-087",
+		models.SkillDef{Name: "helper", Description: "TODO: describe this skill.",
+			Location: models.Location{FilePath: ".claude/skills/helper/SKILL.md"}}, models.RepoInventory{}, true},
+	{"CSKILL-087 silent when the description is real content", "CSKILL-087",
+		models.SkillDef{Name: "helper", Description: "Summarises the current git diff.",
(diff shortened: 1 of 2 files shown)


jhumel-code:
Thanks @kureen-cyber. I compared the fixture copy against the production YAML in trustabl/trustabl-rules#116 and the two are byte identical, which is exactly what `scripts/check-rules-sync.sh` wants, and you correctly left `manifest.yaml` alone since `skill_description_has_text` and `schema_version: 14` are both already on `main`.

The two cases in `policySkillRuleCases` are real rather than vacuous. The predicate lowercases `SkillDef.Description` and does a substring match, so "TODO: describe this skill." fires, and "Summarises the current git diff." contains none of the six needles and stays silent. Field order in the unkeyed literals matches `policySkillCase`, so `TestPolicyRules_AllRulesCovered` will see both a fire and a silent for `CSKILL-087`.

Your read of the `rules-sync` job is right: the ref resolution only looks for branches on `trustabl/trustabl-rules`, not on forks, so it falls back to `main` and reports drift until the rules side lands. I will take trustabl/trustabl-rules#116 first and this one straight after. My notes on the rule itself are on that PR, and a rulebook rationale doc is still needed before the rule reaches users, which we can carry on our side.

sairenchristianbuerano (approved):

PR #213, opened 2026-09-16
test(rules): pin ADK on the TypeScript error-contract predicates

eulestadt:
Follow-up to #169, from the review that landed before merge.

`tsErrorContractCases` said it covered every TypeScript discovery path and listed five. ADK was the missing one, and it is the shape that reads the handler off `execute` rather than a factory argument (`ts_adk_tools.go`), so a fact-walk regression there would not have shown up in the other five.

This adds `google_adk` fire/silent snippets to that table, and updates the two `ARCHITECTURE.md` enumerations that still described the pre-#169 fact set (`ts_handler_facts.go` and the TS-aware predicates list).

`go test ./internal/rules/ -run 'TestPred_HasRaise_TypeScript_Throw|TestPred_HasTryExcept_TypeScript_Caught'` covers the new case.


eulestadt:
Closing this. #169 already merged with an approval, so the review asks were waived. Nothing left to land.

